PR #7, opened 2019-01-21
WIP: Creating prototype of speech HTML parser

mateusbs17:
This speech prototype parser reads a HTML file and create a parsed
object which needs to be stored. The parser is supposed to run when a
new speech is inserted. The parsed speech is then served to
alethei-front to be displayed.


mateusbs17:
The following GIF shows how the HTML looks like and how is the object after parsing.
prototype-aletheia

PR #17, opened 2019-09-03
Repositories Claim and ClaimReview

DuardoN:


packages/server/api/controller/claimController.js
```diff
@@ -1,55 +1,47 @@
...
-        _id: req.params.id
-    }, (err, claim) => {
-        if (err) { res.send(err); }
-        res.json({ message: 'claim successfully deleted' });
-    });
+const ClaimRespository = require('../repository/claim');
+
+module.exports = class ClaimController {
+    listAll() {
+        return new Promise((resolve, reject) => {
```
thesocialdev: Let's use async wait everywhere to keep consistency, what do you think?

packages/server/api/repository/claimReview.js
```diff
@@ -0,0 +1,45 @@
...
+        return ClaimReview.find({}).lean();
+    }
+
+    static create(claimReview) {
+        const newClaimReview = new ClaimReview(claimReview);
+        return newClaimReview.save();
+    }
+
+    static getById(claimReviewId) {
+        return ClaimReview.findById(claimReviewId);
```
thesocialdev: Not populating with claims

packages/server/routes/claimReview.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
-const claimReview = require('../api/controller/claimReviewController');
+const ClaimReview = require('../api/controller/claimReviewController');
```
thesocialdev: For consistency sake, rename ClaimReview to ClaimReviewController. Apply this everywhere for other classes.

DuardoN: ok.

packages/server/api/controller/claimController.js
DuardoN: It's ok. No problem

packages/server/api/repository/claimReview.js
DuardoN: I forgot populate data in repository.

PR #25, opened 2019-10-29
Claim review crud

thesocialdev:
0) Create claim review form and endpoint
1) Create endpoint to get stats given personality id
2) Return stats when fetching personality
3) change paths for claims considering claim as a resource from
personality
4) Stats gives: total of reviews and percentage of review for each label
5) Move docker-compose.yaml to the root dir
6) Fix npm security warnings


packages/server/api/repository/personality.js
```diff
@@ -15,8 +16,47 @@ module.exports = class PersonalityRepository {
...
+    }
+
+    static async getReviewStats(id) {
+        const personality =  await Personality.findById(id);
+        const reviews = await ClaimReview.aggregate([
+            { $match: { personality: personality._id } },
+            { $group: { _id: "$classification", count: { $sum: 1 } } },
+        ]);
+        const total = reviews.reduce((agg, review) => {
+            agg += review.countç;
```
DuardoN: I think the letter **Ç** not exist when you use `count`.

thesocialdev: Good catch!

PR #32, opened 2019-11-09
Personality List

thesocialdev:
1) Fix Grommet center issue
2) Add search input into Personality List and use new endpoint properly
3) Add not found body message
4) Add pagination


packages/front/src/components/PersonalityList.jsx
```diff
@@ -14,30 +27,91 @@ class ClaimCreate extends Component {
...
     getPersonalities() {
-      axios.get('http://localhost:3000/personality')
+      const { page, searchName, pageSize } = this.state;
+      const params = {
+        page,
+        name: searchName,
+        pageSize
+      };
+
+      axios.get('http://localhost:3000/personality', {params})
```
jvzanatta: In the future this can be changed so it uses an environment variable or configuration instead.
https://medium.com/better-programming/using-environment-variables-in-reactjs-9ad9c5322408

thesocialdev: Agreed, there is a task for that https://www.pivotaltracker.com/story/show/169663921

PR #73, opened 2020-02-13
Recaptcha fixes and Wikidata Support

thesocialdev:
When fetching personality that has wikidata ID, get name, description, and image from Wikimedia.


thesocialdev:
If you want to test and create a new personality, you need to do it POSTing to the personality endpoint, here is an example to add Barack Obama as a personality

image

Issue #43, opened 2020-02-09
warning package.json: No license field


thesocialdev:

thesocialdev:
Done.

PR #74, opened 2020-02-13
Introduce parcel and env variables

thesocialdev:
1) Add parcel as bundler and remove unecessary stuf from
create-react-app scripts
2) Start using environment variables
3) Some clean-ups

#38 #61 #43


packages/front/.gitignore
```diff
@@ -17,6 +17,11 @@
...
 .env.production.local
 
+.cache
+dist
+
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+.eslintcache
```
wellcda: @thesocialdev maybe add a newline here? :joy:

packages/front/.env.example
```diff
@@ -1 +1 @@
-URL=http://localhost:3000
\ No newline at end of file
+API_URL=API_URL=http://localhost:3000
```
wellcda: @thesocialdev I think there is a typo here `API_URL=API_URL`

packages/front/.gitignore
thiagobraga: I agree with Wellindinho, you can use [EditorConfig](https://editorconfig.org) to standardize this type of configuration (end of line, indentation).

And you are ignoring `.eslintcache` but it is already versioned. After this PR is accepted, it will be necessary to delete the file and version the deletion to .gitignore works.

packages/front/src/components/Claim/ClaimSentence.css
```diff
@@ -1,15 +1,7 @@
-button.sentence {
+a.sentence {
```
thiagobraga: Maybe instead of specifying the element (`button`, `a`), remove the element and use only the class name, to reduce specificity. And use a semantic name for the class to you know that this class is about a button, like `.btn-sentence`, `.button-sentence`.

packages/front/src/index.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500" />
```
thiagobraga: Maybe you can add before this line a `preconnect` and a `dns-prefetch` to improve the loading of external fonts and avoid the FOUC.

``` html
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
<link rel="dns-prefetch" href="https://fonts.gstatic.com" />
```

packages/front/.gitignore
thesocialdev: EditorConfig + eslint + prettier will come in another patch. Will just put the new line manually this time.

PR #80, opened 2020-03-01
Personality create

thesocialdev:
Page for personality creation

1) Use input search to search wikidata entity for given name


thesocialdev:
This PR is way behind compared to the framework refactor. I need to merge it.

If you find anything worth to fix, please fill an issue.

PR #87, opened 2020-04-23
Support reference

DuardoN:


packages/server/api/repository/supportReference.js
```diff
@@ -0,0 +1,59 @@
...
+                supportReferenceBody
+            );
+            const supportReferenceUpdate = await SupportReference.findByIdAndUpdate(
+                supportReferenceId,
+                newSupportReference,
+                optionsToUpdate
+            );
+            return supportReferenceUpdate;
+        } catch (error) {
+            // TODO: log to service-runner
```
thesocialdev, line 47: Let's start logging everywhere from now on for the next PRs

packages/server/routes/supportReference.js
```diff
@@ -0,0 +1,80 @@
...
+        .delete(req.params.id)
+        .then(result => res.send(result))
+        .catch(error => {
+            next(Requester.internalError(res, error.message));
+        });
+});
+
+module.exports = function(appObj) {
+    return {
+        path: "/supportReference",
```
thesocialdev, line 76: Let's name it "reference" instead of "supportReference". Apply this change everywhere in the code.

packages/server/api/model/supportReference.js
```diff
@@ -0,0 +1,31 @@
+const mongoose = require("mongoose");
```
thesocialdev, line 1: Change this file name to `referenceModel` to follow the naming pattern for models.

packages/server/api/repository/reference.js
```diff
@@ -0,0 +1,54 @@
+const Reference = require("../model/reference");
```
thesocialdev, line 1: change the name here too, the server is failing.

packages/server/api/model/supportReference.js
thesocialdev: Also, claimModel and claimReviewModel should now have a field for a relationship with `reference` table. One claim or claimReview can have many references.

packages/server/api/model/supportReference.js
```diff
@@ -0,0 +1,31 @@
...
+    },
+    citation: {
+        type: String,
+        required: true
+    },
+    type: {
+        type: String,
+        required: true
+    },
+    classification: {
```
thesocialdev, line 17: Classification should an enum:

- Reliable source
- Unreliable source
- Fake

thesocialdev:
@wellcda this is ready for review

PR #100, opened 2020-05-15
Create AletheiaHeader Component

wellcda:
https://www.pivotaltracker.com/story/show/172869592


src/components/Personality/PersonalityView.jsx
```diff
@@ -48,69 +131,41 @@ class PersonalityView extends Component {
 
     render() {
         const personality = this.state.personality;
+        const arraynovo = new Array().fill(100);
```
thesocialdev, line 134: I guess this is not supposed to be here.

wellcda: it isn't, thanks

PR #108, opened 2020-05-27
Footer UI behavior

thesocialdev:
Make footer stick to the bottom


thesocialdev:
@wellcda now it's ready for review. You can also ignore it and add to your branch if you want to or if I did the same as you.

PR #109, opened 2020-05-28
Feature/recaptcha secret

giovannirossini:
Automating ReCaptcha variable replacement

### DIFF
--- a/.github/workflows/gke.yml
+++ b/.github/workflows/gke.yml
@@ -52,7 +52,7 @@ jobs:
       # Build the Docker image
       - name: Build
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/gke.yml
```diff
@@ -39,6 +39,12 @@ jobs:
         run: |
           sed -i "s/ENVIRONMENT/$ENVIRONMENT/g" deployment.yml
 
+      - name: Set ReCaptcha
+        env:
+          RECAPTCHA_SECRET: "{{ secrets.RECAPTCHA_SITEKEY }}"
+        run: |
+          sed -i "s/RECAPTCHA_SECRET/$RECAPTCHA_SECRET/g" config.example.yml
```
thesocialdev, line 46: It should have a previous step where config.example.yml is copied to config.yml and then the config.yml is overwritten. I
need to tweak the config files though, we should have a config.prod.yml file instead of config.example.yml for that purpose.

giovannirossini: The config.example.yml is copied to config.yml after, in a build step in Dockerfile.

thesocialdev:
@giovannirossini feel free to merge if my new commits are fine

PR #120, opened 2020-06-08
Restore reCaptcha server and front

thesocialdev:
1) Remove comments of old code for reCaptcha
2) Check for reCaptcha errors before adding a claimreview in the DB
3) i18n of the ClaimReview form


thesocialdev:
@giovannirossini and @wellcda this needs coordenation when deployed, because it will test the recaptcha changes made in the CD pipeline.

PR #142, opened 2020-11-02
italian translation

paganoromano:


thesocialdev:
@paganoromano nice job!

PR #144, opened 2020-11-05
i do point claimCreate of the task in eng and it

paganoromano:


src/components/Claim/ClaimCreate.jsx
```diff
@@ -24,7 +26,13 @@ class ClaimCreate extends Component {
         if (this.props.edit) {
             axios
                 .get(
-                    `${process.env.API_URL}/claim/${this.props.match.params.claimId}`
+                    `${process.env.API_URL}/claim/${this.props.match.params.claimId}`,
+                    {
```
thesocialdev, line 30: This is not needed, you can remove it.

src/i18n.js
```diff
@@ -62,7 +62,21 @@ const resources = {
             errorMessage: "Error while sending review",
             successMessage: "Review succeeded"
         },
+        claimCreate: {
+            errorMessage: "Error while fetching claim",
+            responseC: "create with success",
+            errorMessage1: "Error while saving claim",
```
thesocialdev, line 68: Avoid to enumerate the filed, be more specific, for example `erroMessage1` should be `errorMessageSave`

src/i18n.js
```diff
@@ -62,7 +62,21 @@ const resources = {
...
+            responceU: "update with success",
+            errorMessage2: "Error while updating claim",
```
thesocialdev, line 70: same as above

src/i18n.js
```diff
@@ -62,7 +62,21 @@ const resources = {
...
+            message: "Please insert a title",
+            placeholder: "Some title",
+            message1: "Please insert the content",
+            placeholder1: "claim"
```
thesocialdev, line 74: Use the capital letters in the beginning always.

src/components/Claim/ClaimCreate.jsx
```diff
@@ -1,9 +1,11 @@
+import { withTranslation } from "react-i18next";
 import React, { Component } from "react";
 import axios from "axios";
 import { Editor, EditorState, ContentState } from "draft-js";
 import { stateToHTML } from "draft-js-export-html";
 import { stateFromHTML } from "draft-js-import-html";
 import "draft-js/dist/Draft.css";
+
```
thesocialdev, line 8: Please don't add a new line between imports

thesocialdev (changes requested):

PR #145, opened 2020-11-12
Make log accessible from repository interface

jvzanatta:


thesocialdev:
@jvzanatta can you take a look at the SonarCloud suggestions?

PR #151, opened 2021-03-13
Implement login and finish Login Page

thesocialdev:
1) Tweak server to authenticate requests properly using cookies session
2) Create simple login page
3) Add user/validate endpoint to validate the user session
4) Refactor store reducer to include login state property alongside search
5) Implement client-side session validate when app is mounted


thesocialdev:
CI is complaining about code duplication in the i18n files. I'll merge it knowing that this will be revisited in the future.

PR #189, opened 2021-05-29
Homepage UI

julianoebram:
Style for see more personalities button. Add arrow right icon.

### DIFF
--- a/src/components/Home/Home.jsx
+++ b/src/components/Home/Home.jsx
@@ -6,6 +6,7 @@ import statsApi from "../../api/stats";
 import PersonalityCard from "../Personality/PersonalityCard";
 import { withTranslation } from "react-i18next";
 import SocialMediaShare from "../SocialMediaShare";
+import { ArrowRightOutlined } from "@ant-design/icons"
 
 class Home extends Component {
     constructor(props) {
@@ -190,8 +191,13 @@ class Home extends Component {
                                 padding: "10px"
                             }}
                         >
-                            <Button href="/personality">
+                            <Button href="/personality" style={{
+                                border: "2px solid #2D77A3",
+                                boxSizing: "borderBox",
+                                borderRadius: "30px",
+                            }}>
                                 {t("home:seeMorePersonalitiesButton")}
+                                <ArrowRightOutlined />
                             </Button>
                         </Row>
                     </Row>
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 .env.production.local
 .env
 .eslintcache
+.vscode/**/*
 
 npm-debug.log*
 yarn-debug.log*


src/components/Personality/PersonalityCard.jsx
```diff
@@ -99,6 +126,27 @@ class PersonalityCard extends Component {
...
+
+                                            height: "3em",
+                                            width: "8em",
+                                            paddingTop: "1em",
+
+                                            fontFamily: "Open Sans",
+                                            fontStyle: "normal",
+                                            fontWeight: "bold",
+                                            fontSize: "14px",
+                                            lineHeight: "107%",
```
thesocialdev, line 144: This is not needed. If you remove that you don't need to change width, height, and paddingTop.

src/components/Personality/PersonalityCard.jsx
```diff
@@ -99,6 +126,27 @@ class PersonalityCard extends Component {
                                         }
                                         href={`${this.props.hrefBase ||
                                             "personality/"}${personality._id}`}
+
+                                        style={{
+                                            background: "#2D77A3",
+                                            borderRadius: "30px",
+
+                                            position: "absolute",
```
thesocialdev, line 134: Not needed

src/components/Personality/PersonalityCard.jsx
```diff
@@ -99,6 +126,27 @@ class PersonalityCard extends Component {
...
+
+                                            height: "3em",
+                                            width: "8em",
```
thesocialdev, line 137: Width should be 100%, we don't need to have the exact same dimensions as the Figma mockup, just the proportions.

src/components/Personality/PersonalityCard.jsx
```diff
@@ -99,6 +126,27 @@ class PersonalityCard extends Component {
                                         }
                                         href={`${this.props.hrefBase ||
                                             "personality/"}${personality._id}`}
+
+                                        style={{
+                                            background: "#2D77A3",
```
thesocialdev, line 131: In this case, you need to change the button's border color as well. It's using the default color.

src/components/Personality/PersonalityCard.jsx
```diff
@@ -33,12 +33,39 @@ class PersonalityCard extends Component {
...
-                            marginTop: "10px",
-                            width: "100%"
+                            marginTop: "1em",
+
+                            width: "1000px",
+
+                            position: "relative",
+                            height: "156px",
+                            marginLeft: "-4.95%",
+                            marginRight: "-4%",
```
thesocialdev, line 43: We should avoid negative margins, if we need to do this, something is not right with the HTML structure.

src/components/Home/Home.jsx
```diff
@@ -188,10 +220,38 @@ class Home extends Component {
...
+                                alignItems: "center",
+                                justifyContent: "center",
+
+                                border: "2px solid #2D77A3",
+                                boxSizing: "borderBox",
+                                borderRadius: "30px",
+                                height: "40px",
+                                color: "#2D77A3",
+
+                                fontFamily: "Open Sans",
```
thesocialdev, line 243: Since it's repeating a lot, maybe we should move font-family and font-style to the body CSS

src/components/Home/Home.jsx
```diff
@@ -188,10 +220,38 @@ class Home extends Component {
...
+                                fontStyle: "normal",
+                                fontWeight: "bold",
+                                fontSize: "14px",
+                                lineHeight: "107%",
```
thesocialdev, line 247: Again, this is not needed and can remove other CSS rules being used to fix what this is causing.

src/components/Home/Home.jsx
```diff
@@ -167,8 +169,38 @@ class Home extends Component {
...
+                                        width: "150px",
+                                        height: "40px",
+                                        padding: "12px",
+
+
+                                        fontFamily: "Open Sans",
+                                        fontStyle: "normal",
+                                        fontWeight: "bold",
+                                        fontSize: "14px",
+                                        lineHeight: "107%",
```
thesocialdev, line 198: Same here.

src/components/Home/Home.jsx
```diff
@@ -167,8 +169,38 @@ class Home extends Component {
                                     justifyContent: "space-between"
                                 }}
                             >
-                                <span>{t("home:statsFooter")}</span>
-                                <Button href="#create_account">
+                                <span style={{
+                                    position: "relative",
+                                    width: "133px",
```
thesocialdev, line 174: This will not be responsive if you set fixed width values

src/components/Home/Home.jsx
```diff
@@ -167,8 +169,38 @@ class Home extends Component {
...
+                                    height: "2em",
+                                }}
+                                >{t("home:statsFooter")}</span>
+                                <Button href="#create_account"
+                                    style={{
+                                        position: "absolut",
```
thesocialdev, line 180: There is a type "absolute". But this rule is not needed as well.

src/components/Home/Home.jsx
```diff
@@ -188,10 +220,38 @@ class Home extends Component {
...
-                            <Button href="/personality">
+                            <Button href="/personality" style={{
+
+
+                                margin: "auto",
+                                left: "30%",
+                                bottom: "10px",
+                                position: "relative",
+
+                                alignItems: "center",
```
thesocialdev, line 234: Has no effect

src/components/Home/Home.jsx
```diff
@@ -188,10 +220,38 @@ class Home extends Component {
...
+                                justifyContent: "center",
```
thesocialdev, line 235: Has no effect

src/components/Personality/PersonalityCard.jsx
```diff
@@ -33,12 +33,39 @@ class PersonalityCard extends Component {
                     <Row
                         style={{
                             padding: "10px 30px",
-                            marginTop: "10px",
-                            width: "100%"
+                            marginTop: "1em",
+
+                            width: "1000px",
```
thesocialdev, line 38: This will not be responsive

src/components/Personality/PersonalityCard.jsx
```diff
@@ -33,12 +33,39 @@ class PersonalityCard extends Component {
...
+
+                            position: "relative",
+                            height: "156px",
```
thesocialdev, line 41: The card should adapt the height according to it's content, this is not needed.

thesocialdev (changes requested):

PR #196, opened 2021-08-20
Migrate App Home page to nextjs

thesocialdev:


thesocialdev:
@DuardoN and @jvzanatta, this is an example of one page (Home) fully migrated to nextjs. We should merge this before you engage with the work for the other pages.

PR #200, opened 2021-09-20
179298695/migrate personality page

jvzanatta:


src/components/List/BaseList.tsx
```diff
@@ -0,0 +1,127 @@
...
+        apiCall(query).then(newItems => {
+            setInitLoading(false)
+            setLoading(false)
+            setTotalPages(newItems.totalPages)
+            setTotalItems(newItems.total)
+            setItems(items => [...items, ...newItems.data])
+        });
+    }, [query, apiCall]);
+
+    const loadMoreData = () => {
```
thesocialdev, line 37: This seems to be broken, it doesn't make the request to page to the next items of the list

thesocialdev (approved):
There are a few follow-up fixes that need to be added in a follow-up patch.

PR #270, opened 2022-02-04
Add new Report field to ClaimReview

thesocialdev:
- Bump nextjs packages versions
- Remove unused packages
- Fix some wrong imports from react-i18next
- Fix virtual populate not working with aggregate function
- Show sources and report in the ClaimReviewCard component


thesocialdev:
Merging anyway, since we don't care for 80% code coverage at the moment.

PR #271, opened 2022-02-04
Release Beta: Add new Report field to ClaimReview

thesocialdev:


thesocialdev:
Merging anyway, since we don't care for 80% code coverage at the moment.

PR #318, opened 2022-02-24
Fixed input fields color from lightgray to white(FFF, White)

caneppelevitor:


src/components/Input.tsx
```diff
@@ -3,7 +3,7 @@ import { Input } from "antd";
 
 
 const AletheiaInput = styled(Input)`
-    background: #F5F5F5;
+    background: #FFF;
```
thesocialdev, line 6: Let's use the colors library here as well

thesocialdev (changes requested):

PR #335, opened 2022-03-02
Created a date component with toLocaleDateString

caneppelevitor:


src/components/LocalizedDate.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react"
+
+const LocalizedDate = ( {date} ) => {
+  const localizedDate = new Date(date).toLocaleDateString() 
+return (
+  <>
+  <strong>{localizedDate}</strong>&nbsp;
```
thesocialdev, line 7: let's remove the `&nbsp;`, it's not a requirement for the LocalizedDate component and it's only needed in the ClaimCardHeader

src/components/LocalizedDate.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react"
+
+const LocalizedDate = ( {date} ) => {
+  const localizedDate = new Date(date).toLocaleDateString() 
+return (
+  <>
```
thesocialdev, line 6: if you remove `&nbsp;` you don't need `<></>`

thesocialdev (changes requested):

thesocialdev (approved):

PR #342, opened 2022-03-09
Ory auth pages#337

### LINKED ISSUE #274: Replace user Authentication flow with Ory Kratos
## The problem statement
The user authentication handling might be too difficult for us to tackle. The idea here is to move away from Passport.js and start using a vendor that we can integrate seemingly. There are options out there, like Auth0.

But, we need to user OSS complaint software as much as possible, therefore, for this task we should be using [Ory Kratos](https://ory.sh).

## Open questions
- [x] #297
- [x] #296

## Acceptance Criteria
- [x] #337
- [x] #338
- [x] The user flow authentication uses Ory Kratos
- [x] ~~The old user logic is cleaned-up~~
- [ ] The new Ory flow authentication is successfully integrated with current functionality that depends on users
- [x] The current user seed capability is update to the new auth flow

### LINKED ISSUE #337: Create Login and Sign up page in the client-side
## Background information
Now that we know Ory can be used after #297. We need to implement it.

Because Ory creates cookie sessions during login, we do not need the server-side to authenticate the Login page, let's create it first.

OBS: this will break Auth functionality and server-side AuthGuard need to be implemented as well

## Acceptance Criteria
- [x] We can create an user in Aletheia using ORY interfaces using user seed 2ba7eca589e19c90e22500cf4971bb6bda087ff3
- [x] We can login the user with ORY and submit a request that contains the proper authentication cookies

### LINKED ISSUE #338: Update Auth Guard in the server-side to work with Ory

ThaisMap:
- [x] Installed Ory Kratos Client and created necessary files 
- [x] Create user with user seed
- [x] Login page
- [x] AuthGuard
- [x] Query proper user id when authenticating with Ory

Closes #337
Closes #338 
Closes #274


thesocialdev:
So, it turns out login wasn't working at all and there's some problem with compatibility with the `ory/integrations` library. I've registered an issue upstream to get some help to fix it.

Blocked on https://github.com/ory/integrations/issues/29

PR #343, opened 2022-03-11
Introduce Claim Revision with sequence diagrams

pepermao:
- [x] Claim creation
- [x] Claim update


server/claim/claim.service.ts
```diff
@@ -38,7 +38,6 @@ export class ClaimService {
             .sort({ _id: order })
             .lean();
         return Promise.all(
-            /** Get lastest revision for each claim */
             claims.map(async (claim) => {
```
thesocialdev, line 41: Here's a place where we forgot to update with the new `this._getClaim` functionality.

server/personality/personality.service.ts
```diff
@@ -95,7 +93,7 @@ export class PersonalityService {
         personality.claims = await Promise.all(personality.claims.map((claim) => {
             return {
                 ...claim,
-                ...claim.revisions[0]
+                ...claim.lastestRevision
```
thesocialdev, line 96: we should clean-up the query above, although this is working, we don't need to populate revisions the way we did before

thesocialdev (changes requested):

server/claim/claim.service.ts
pepermao: I didn't understand

server/claim/claim.service.ts
```diff
@@ -46,80 +54,78 @@ export class ClaimService {
...
-
-        return this.postProcess(claim.toObject());
+        return this._getClaim({_id: claimId})
     }
 
     async getByPersonalityIdAndClaimSlug(personalityId, claimSlug) {
+        return this._getClaim({personality: personalityId, slug: claimSlug})
+    }
+
+    private async _getClaim(match: ClaimMatchParameters, postprocess = true) {
```
thesocialdev, line 101: @pepermao I meant that we have created this new function that retrieves Claim with a revision, and execute the postProcess function.

In the comment above we should use _getClaim and refactor the functionality.

thesocialdev: But let's do this in a another PR because it's going to be a little tricky.

thesocialdev (approved):

PR #352, opened 2022-03-23
change the way we get the language using middleware (refactoring)

pepermao:
- Refactoring code. Now we can get the language with middleware, which does the same thing as before, but with less code


server/app.module.ts
```diff
@@ -60,4 +66,16 @@ export class AppModule {
             ],
         };
     }
+
+    configure(consumer: MiddlewareConsumer) {
+        consumer
+            .apply(GetLanguageMiddleware)
```
thesocialdev, line 72: Can we always apply the middleware to all routes?

thesocialdev (changes requested):

thesocialdev (approved):

Issue #323, opened 2022-02-28
Server | Create a History Module that tracks creation and update between revisions for Claim and Personality


thesocialdev:
### Background Information
In order to track history in the platform, we need to register every time creation and an update is performed.

### How
We need to have a "before" and "after" state for the specific DB Document.

#### For Personality
Because personality doesn't have revisions we should store the current Personality state as an object for the "before" entry and the updated state for the "after".

When it's created the "before" entry is empty.

### Suggested Schema
```
{
  targetId: uuid, // personality id or claim id
  type: Enum<["create", "update", "delete"]>
  before: any
  after: any
}
```

#### For Claim
We just need to set the previous revision and the current revision.

### Acceptance Criteria
- [x] When creating and updating Claim and Personalities History crates a new entry
- [x] Update documentation

PR #365, opened 2022-03-31
#323  Create a History Module that tracks creation and update between revisions for Claim and Personality

pepermao:
So far, we can track changes to claims when creating and updating

### DIFF
--- a/server/claim/claim.service.ts
+++ b/server/claim/claim.service.ts
@@ -1,4 +1,4 @@
-import {Injectable, Logger} from "@nestjs/common";
+import { Injectable, Inject, Logger, Scope } from "@nestjs/common";
 import { InjectModel } from "@nestjs/mongoose";
 import { FilterQuery, Model, Types } from "mongoose";
 import { Claim, ClaimDocument } from "../claim/schemas/claim.schema";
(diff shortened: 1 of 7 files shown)


server/history/schema/history.schema.ts
```diff
@@ -0,0 +1,44 @@
...
+    @Prop({
+      type: Object,
+      required: true,
+    })
+    details: object
+}
+
+export const HistorySchema = SchemaFactory.createForClass(History);
+
+//ter uma validação dos tipos, caso seja um tipo create, é obrigatório um after
```
thesocialdev, line 44: That's a TODO for another patch? Could you write it in English, please?

server/claim/claim.service.ts
```diff
@@ -78,6 +106,10 @@ export class ClaimService {
             })
         claim.latestRevision = newClaimRevision._id
         claim.slug = newClaimRevision.slug
+        
+        const history = this._getHistoryParams(claimId, this.req.user, HistoryType.Update, newClaimRevision, latestRevision)
```
thesocialdev, line 110: Let's rename `latestRevision` to `previousRevision` to be more semantic.

server/history/schema/history.schema.ts
pepermao: I think i'll resolve in this patch, just to reminder myself

thesocialdev (changes requested):

server/claim-revision/claim-revision.service.ts
```diff
@@ -23,16 +23,10 @@ export class ClaimRevisionService {
...
     getRevision(claimId) {
         return this.ClaimRevisionModel.findOne({ claimId })
     }
-    /**
-     *
-     * @param claimId an unique claim id
-     * @param claim Claim Content
-     * @returns Save the claimRevision in database
-     */
+
```
thesocialdev, line 29: Any reason for removing this documentation?

thesocialdev (approved):

PR #369, opened 2022-04-04
Add to the history deleted Claim and Personality

pepermao:
- [x] History Delete of claims
- [x] History Delete of personalities


server/claim/claim.controller.ts
```diff
@@ -248,26 +248,31 @@ export class ClaimController {
...
+
+            await this.viewService
+                .getNextServer()
+                .render(
+                    req,
+                    res,
+                    "/claim-page",
+                    Object.assign(parsedUrl.query, { personality, claim })
+                );
+        } catch {
```
thesocialdev, line 271: Nice, you find a way of solving it!

I would suggest that, instead of executing next server to render the 404 page, you should raise a NotFoundException and let the [exception filter execute a redirect to 404.](https://github.com/AletheiaFact/aletheia/blob/develop/server/filters/not-found.filter.ts#L7)

Also, you should raise the exception in the service, not in the controller.

server/personality/personality.controller.ts
```diff
@@ -105,20 +105,25 @@ export class PersonalityController {
...
             );
+            await this.viewService
+                .getNextServer()
+                .render(
+                    req,
+                    res,
+                    "/personality-page",
+                    Object.assign(parsedUrl.query, { personality })
+                );
+        } catch {
```
thesocialdev, line 122: Same as above

thesocialdev (changes requested):

thesocialdev (approved):

PR #375, opened 2022-04-06
implement history module when create and delete a claim review

pepermao:
image
 
image

image

when soft-delete, although 'isDelete' is equal to true, the claim revision still with the revision


server/claim-review/claim-review.service.ts
```diff
@@ -165,8 +166,16 @@ export class ClaimReviewService {
...
 
+    /**
+     * This function creates a new claim review.
+     * Also creates a History Module that tracks creation of claim reviews.
+     * @param claimReview ClaimReviewBody received of the client.
+     * @returns Return a new claim review object.
+     */
     async create(claimReview) {
         // Cast ObjectId
+        //verificar id da claim review
```
thesocialdev, line 177: Right, you can remove this comment.

server/history/schema/history.schema.ts
```diff
@@ -7,6 +7,7 @@ export type HistoryDocument = History & mongoose.Document;
 export enum TargetModel { 
   Claim = 'claim',
   Personality = 'personality',
+  ClaimReview = 'claim-review'
```
thesocialdev, line 10: Use pascal case for the TargetModel enum

PR #376, opened 2022-04-06
changed the claim revision id to claim revision object in history module

pepermao:
when create a claim, param after is changed to the claim revision object, not the claim revision id

### DIFF
--- a/server/claim/claim.service.ts
+++ b/server/claim/claim.service.ts
@@ -77,7 +77,7 @@ export class ClaimService {
                 TargetModel.Claim,
                 user,
                 HistoryType.Create,
-                newClaim.latestRevision
+                newClaimRevision
             )
         await this.historyService.createHistory(history)
 


server/claim/claim.service.ts
```diff
@@ -77,7 +77,7 @@ export class ClaimService {
                 TargetModel.Claim,
                 user,
                 HistoryType.Create,
-                newClaim.latestRevision
```
thesocialdev, line 80: If we do that we start to duplicate data unnecessarily, why do we need to do that?

thesocialdev (changes requested):

server/claim/claim.service.ts
thesocialdev: As discussed in slack, let's keep the revision id as the `before` and `after` content and when requesting the specific history entry, we will do an extra query to pull the content from the revisions cc/ @ThaisMap

PR #384, opened 2022-04-08
Introduce ClaimRevisionById Get Function.

caneppelevitor:
### Problems who we run into  #392 #393


server/claim/claim.service.ts
```diff
@@ -182,7 +192,8 @@ export class ClaimService {
     private async postProcess(claim) {
         // TODO: we should not transform the object in this function
         claim = {
-            ...claim.latestRevision, // the order matters
+            ...claim?.latestRevision,// the order matters
```
thesocialdev, line 195: Change to one line as:
```
...(claim?.latestRevision || claim?.revision)
```

server/claim/claim.service.ts
```diff
@@ -154,24 +154,34 @@ export class ClaimService {
...
     }
 
-    async getByPersonalityIdAndClaimSlug(personalityId, claimSlug) {
-        return this._getClaim({personality: personalityId, slug: claimSlug})
+    async getByPersonalityIdAndClaimSlug(personalityId, claimSlug, revisionId = undefined) {
+        return this._getClaim({personality: personalityId, slug: claimSlug}, revisionId)
     }
 
-    private async _getClaim(match: ClaimMatchParameters, postprocess = true) {
+    private async _getClaim(match: ClaimMatchParameters, revisionId = undefined, postprocess = true) {
```
thesocialdev, line 161: The post-process flag is not a good idea, we need to manipulate the claim object properly so it will have the same structure when `this.postProcess` is called

thesocialdev: My suggestion is something like:
```
        let claim
        if(revisionId) {
            claim = {
                ...await this.ClaimModel.aggregate([
                    { $match: match },
                    { $project: { "latestRevision": 0}}
                ])[0],
                revision: await this.claimRevisionService.getRevision(revisionId)
            }
        } else {
           claim = await this.ClaimModel.findOne(match)
                .populate("personality", "_id name")
                .populate("sources", "_id link classification")
                .populate("latestRevision").toObject()
        }

        if (!claim) {
              throw new NotFoundException()
        }
        return this.postProcess(claim)
```

server/claim-revision/claim-revision-controller.ts
```diff
@@ -0,0 +1,12 @@
+import { Controller, Get, Param } from "@nestjs/common";
```
thesocialdev, line 1: Is the route really being used?

server/claim-revision/claim-revision.module.ts
```diff
@@ -25,6 +26,7 @@ const ClaimRevisionModel = MongooseModule.forFeature([
         ViewModule,
         SourceModule,
     ],
+    controllers: [ClaimRevisionController],
```
thesocialdev, line 29: I don't think we need this route for now, let's remove it.

thesocialdev (changes requested):

server/claim/claim.service.ts
pepermao: if we remove the conditional  return postprocess === true ? this.postProcess(claim.toObject()) : claim;

we'll affect the update function, which doesn't need to pass in postProcess function

pepermao: and the declaration:
...await this.ClaimModel.aggregate([
                    { $match: match },
                    { $project: { "latestRevision": 0}}
                ])[0],
doesn't returns the claim

pepermao: image
 i had to make this

pepermao: its having a lot of conflicts let's make a call

thesocialdev (approved):

PR #388, opened 2022-04-11
fixed create a deleted personality error

### LINKED ISSUE #387: Server | When create a personality check first if it has already been deleted
### Background Information
When create a personality that has already been deleted, just change the isDeleted to false, this way we avoid a wikitadaId duplicate error.

pepermao:
Closes #387

### DIFF
--- a/server/personality/personality.service.ts
+++ b/server/personality/personality.service.ts
@@ -79,31 +79,49 @@ export class PersonalityService {
      * @param personality PersonalityBody received of the client.
      * @returns Return a new personality.
      */
-    create(personality) {
+    async create(personality) {
         try {
-            personality.slug = slugify(personality.name, {
-                lower: true,     // convert to lower case, defaults to `false`
-                strict: true     // strip special characters except replacement, defaults to `false`
-            })
-            const newPersonality = new this.PersonalityModel(personality);
-            this.logger.log(
-                `Attempting to create new personality with data ${personality}`
-            );
-
-            const user = this.req.user
-
-            const history = this.history.getHistoryParams(
-                newPersonality._id,
-                TargetModel.Personality,
-                user,
-                HistoryType.Create,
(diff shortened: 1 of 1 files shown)


server/personality/personality.service.ts
```diff
@@ -79,33 +79,51 @@ export class PersonalityService {
...
-                HistoryType.Create,
-                personality
-            )
-
-            this.history.createHistory(history)
-
-            return newPersonality.save();
+            const personalityExisted = await this.getDeletedPersonalityByWikidata(personality.wikidata)
+            if(personalityExisted) {
+                return personalityExisted
```
thesocialdev, line 86: Do the restore in this line of code, to decouple the procedure from the `get` operation. I mean that the `getDeletedPersonalityByWikidata` should just return the personality if it exists and is deleted and not restored.

server/personality/personality.service.ts
```diff
@@ -79,33 +79,51 @@ export class PersonalityService {
...
+    
+                this.history.createHistory(history)
+    
+                return newPersonality.save();
+            }
         } catch (err) {}
     }
 
+    async getDeletedPersonalityByWikidata(wikidata) {
+        const deletedPersonalities = await this.PersonalityModel.findDeleted(true)
```
thesocialdev, line 115: What do you think about executing a find and pass the wikidata as a paramter? Like that:
```
this.personalityModel.find({ wikidata, deleted: true})
```

That way you don't need to filter the array afterwards. The wrong choice here is to get all deleted and iterate over them, what if you have 1 million deleted personalities? This would cause a performance penalty and this endpoint could break or become very slow. Furthermore, this could also become a DDOS breach.

thesocialdev (changes requested):

thesocialdev (approved):

PR #399, opened 2022-04-14
Start using next seo in claim page#307

pepermao:
https://seoanalyzer.me/resources/pdf-reports/ea4b-138-97-120-243.sa.ngrok.io_438059.pdf


src/pages/claim-page.tsx
```diff
@@ -1,11 +1,53 @@
...
+            }
+        })
+    })
+
+    
     return (
-        <ClaimView personality={personality} claim={claim} href={href}/>
+        <>
+            <Head>
+                <script type="application/ld+json">{`
```
thesocialdev, line 27: Nice! But let's do this on a separate PR and use another issue for it. I have a lot of considerations for this feature.

server/root/root.controller.ts
```diff
@@ -1,6 +1,6 @@
 import { Controller, Get, Req, Res } from "@nestjs/common";
 
-@Controller("api")
+@Controller()
```
thesocialdev, line 3: The health endpoint should still have the `api` prefix, otherwise it will break how `kubernetes` check for the service health in production.

src/components/Personality/PersonalityCard.tsx
```diff
@@ -49,7 +49,7 @@ const PersonalityCard = ({
 
     if (personality) {
         return (
-            <Row
+            <section
```
thesocialdev, line 52: Same as above

public/locales/pt/seo.json
```diff
@@ -2,5 +2,10 @@
     "personalityDescription": "Veja o perfil de {{name}} na AletheiaFact.org",
     "personalityImageAlt": "Imagem de {{name}}",
     "personalityListTitle": "Perfis de personalidades",
-    "personalityListDescription": "Veja todos os perfis na AletheiaFact.org"
+    "personalityListDescription": "Veja todos os perfis na AletheiaFact.org",
```
thesocialdev, line 5: Missing English versions?

thesocialdev (changes requested):

server/root/root.controller.ts
pepermao: but with api prefix we run into SEO problem

image

thesocialdev: Let's put the prefix only for the health endpoint

PR #405, opened 2022-04-15
start Improvement of share components

### LINKED ISSUE #308: Improve share components to allow backlink to grow organic traffic
### How
- [x] Always set the AletheiaFact.org logo in the share images
- [x] Track clicks to the with umami analytics

### Open Questions
- [x] How to set hashtags for the social links?
R: using hashtag props of react-share component

- [x] How to tag personalities?
R: Getting social media usernames of each personality and put an @ before the username. This works in instagram, twitter and facebook

- [ ] Can we make sure the share was completed?
R: i didn't know exactly, but i found two discussions on stack overflow
https://stackoverflow.com/questions/8748501/checking-if-someone-has-shared-a-certain-link-on-facebook
https://stackoverflow.com/questions/27673948/how-do-i-know-when-someone-has-shared-my-link-on-social-media

pepermao:
image

Closes #308

### DIFF
--- a/server/personality/dto/create-personality.dto.ts
+++ b/server/personality/dto/create-personality.dto.ts
@@ -11,6 +11,18 @@ export class CreatePersonality {
     @IsString()
     wikidata: string;
 
+    @IsString()
+    @IsOptional()
+    P2013?: string;
+
+    @IsString()
+    @IsOptional()
+    instagramUsername?: string;
+
+    @IsString()
+    @IsOptional()
+    P2002?: string;
+
     @IsBoolean()
     @IsOptional()
     isAllowedProp?: boolean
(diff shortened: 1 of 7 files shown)


server/personality/dto/create-personality.dto.ts
```diff
@@ -11,6 +11,18 @@ export class CreatePersonality {
     @IsString()
     wikidata: string;
 
+    @IsString()
```
thesocialdev, line 14: As discussed in Slack. Let's remove the Wikidata integration for now. The idea is to just confirm that we can tag a personality and we can implement that in a follow-up patch.

thesocialdev (changes requested):

thesocialdev (approved):

PR #473, opened 2022-05-24
show only claim reviews published

pepermao:
## In this picture we did a review, but it's not published
image

## In this picture we did a review and its publised
image
## See that only the stats of the published review appear
image

image

closes #473


thesocialdev:
@pepermao this PR should be merged with at least the backend functionality to save the claim review as draft

thesocialdev:
@pepermao should we rebase it and prepare for merging?

PR #476, opened 2022-05-30
Account Settings Integrate With Ory 

caneppelevitor:


server/users/users.controller.ts
```diff
@@ -83,16 +71,17 @@ export class UsersController {
     }
 
     @Get("profile")
-    public async profile(@Req() req: Request, @Res() res: Response) {
+    public async profile(@Req() req, @Res() res: Response) {
```
thesocialdev, line 74: any reason for removing the type hint here?

src/components/Login/OryLoginForm.tsx
```diff
@@ -23,8 +23,12 @@ const OryLoginForm = () => {
         oryGetLoginFlow({ router, setFlow, t })
     }, [])
 
+    useEffect(() => {
+        console.log(flow)
```
thesocialdev, line 27: remember to remove all console.logs

thesocialdev (changes requested):

server/users/users.controller.ts
ThaisMap: The typing was not recognizing the req.user._id

thesocialdev: What do you think about adding a new type? Maybe something like:
```
type BaseRequest = Request & { user: { _id: string }}
```

PR #480, opened 2022-06-08
Workflow prototype frontend

ThaisMap:
Style fields to match current identity
Create a type for the form fields used in the dynamic form
Apply i18n

### DIFF
--- a/src/components/InputTextList.tsx
+++ b/src/components/InputTextList.tsx
@@ -3,16 +3,23 @@ import * as React from "react";
 import { useForm, useFieldArray } from "react-hook-form";
 import Button from "./Button";
 import { Col, Row } from "antd";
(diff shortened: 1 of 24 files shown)


src/components/InputTextList.tsx
```diff
@@ -80,7 +82,7 @@ export default function InputTextList(props) {
                         })
                     }
                 >
-                    <PlusOutlined /> Add another input
+                    <PlusOutlined /> {t("claimReviewForm:addInput")}
```
thesocialdev, line 85: I think it would be useful to move the i18n string to the prop and create a prop `addButtonLabel` or something like that

thesocialdev (changes requested):

thesocialdev (approved):

PR #491, opened 2022-06-13
Update the seed script to work with Ory

### LINKED ISSUE #490: Update seed script to work with Ory
## Background Information
We currently get an error when trying to seed users that already exist in Ory. We need to update the seed.ts script to make it work when we already have the user in Ory, but not in our database.

## Suggestion
Manually get the user_id in the Ory console and include it in config.seed.yaml and if the script finds this prop don't try to create the ory identity and just save the user in the database.

ThaisMap:
if the script finds an oryId prop in the config.seed.yaml it wont try to create the ory identity and just save the user in the database.

Closes #490

### DIFF
--- a/server/users/users.service.ts
+++ b/server/users/users.service.ts
@@ -1,8 +1,9 @@
-import { Model } from "mongoose";
-import {Injectable, Logger} from "@nestjs/common";
-import { InjectModel } from "@nestjs/mongoose";
-import { User, UserDocument } from "./schemas/user.schema";
-import OryService from "../ory/ory.service";
+import { Injectable, Logger } from '@nestjs/common';
+import { InjectModel } from '@nestjs/mongoose';
+import { Model } from 'mongoose';
+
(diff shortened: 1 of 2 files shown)


server/users/users.service.ts
```diff
@@ -13,25 +14,30 @@ export class UsersService {
...
             );
+            newUser.oryId = oryUser.id;
+        }
+        try {
+            // @ts-ignore
+            return this.UserModel.register(newUser, user.password);
         } catch (e) {
-            await this.oryService.deleteIdentity(oryUser.id)
+            if (!newUser.oryId) {
+                await this.oryService.deleteIdentity(newUser.oryId);
```
thesocialdev, line 39: We can stop doing this considering that we will be creating some users outside the platform through Ory Cloud.

thesocialdev (changes requested):

thesocialdev (approved):

PR #492, opened 2022-06-14
save a review as draft before publishing

### LINKED ISSUE #485: Make it possible to save a review as draft before publishing

pepermao:
closes #485


src/utils/ParseMachineState.ts
```diff
@@ -1,4 +1,7 @@
 export const ParseMachineState = ({ value, context }) => {
+    if(value !== "published") {
+        value = Object.keys(value)[0]
```
thesocialdev, line 3: Add a comment here to explain why you need to get the first key and why value is an Object rather than a string.

Additionaly, you should add a check wether value is an Object before extracting the key.

thesocialdev (changes requested):

src/utils/ParseMachineState.ts
pepermao: The states assigned and reported has states compound, and when we want to save the machine value, we get "assigned: undraft" for example or "reported: draft". If we save the value in this way, when we try to persiste state, the machine don't recognized the initial value 'cause the value it's a object. So for these states i get only the key value to save in dataBase, and for others states i save normally

thesocialdev: I got it, please add a comment in the code for posteriority.

src/machine/reviewTaskMachine.ts
```diff
@@ -54,10 +92,13 @@ export const transitionHandler = (state) => {
                 api.createClaimReviewTask({ sentence_hash, machine: state }, t)
                 break;
             case ReviewTaskEvents.finishReport:
```
thesocialdev, line 94:
```
        switch (state.event.type) {
            case ReviewTaskEvents.assignUser:
                api.createClaimReviewTask({ sentence_hash, machine: state }, t)
                break;
            default:
                api.updateClaimReviewTask({ sentence_hash, machine: state }, t, state.event.type)
                break;
        }}
```

thesocialdev (changes requested):

thesocialdev (approved):

PR #495, opened 2022-06-17
Isolate recaptcha logic

### LINKED ISSUE #494: Isolate reCaptcha logic

ThaisMap:
We Isolated recaptcha validation in a separate component and a separate endpoint and removed that logic from claim create and claim review form

Closes #494


src/components/Claim/ClaimCreate.tsx
```diff
@@ -122,20 +107,10 @@ const ClaimCreate = ({ personality, claim = {}, sitekey, edit = false }) => {
...
-            setDisableSubmit(true);
-            resolve();
-        });
-    }
 
-    const onChangeCaptcha = () => {
-        const recaptchaString = recaptchaRef.current.getValue();
-        setRecaptcha(recaptchaString);
+    const onChangeCaptcha = (captchaValid) => {
+        setDisableSubmit(!captchaValid)
```
thesocialdev, line 112: I'm wondering if the is going to be secure. If we do not check during the claim submit we open a vulnerability for the claim to be created by simply sending a Boolean value in the request. I think we can separate the logic but not create a new endpoint for that.

thesocialdev (changes requested):

src/components/Claim/ClaimCreate.tsx
thesocialdev: @pepermao can help with that if needed, but I think the best solution would be to create a [ValidationPipe](https://docs.nestjs.com/techniques/validation#using-the-built-in-validationpipe) for the DTO that does the recaptcha check.

ThaisMap: Considering the problems we had with injecting configService in a validationPipe, see if we can use the captcha service approach that I tried on next commit.

thesocialdev:
@ThaisMap we need to rebase it now, but I'm ready to approve it

thesocialdev (approved):

PR #499, opened 2022-06-22
Test e2e working with cypress run

Keikonichi:
### Background image

Test e2e working with recaptcha in the right order with cypress run

image


Keikonichi:
As for the reCAPTCHA keys that are used in the tests, they will be inserted into the individual .env 
Should these keys be inserted in this pull request?

PR #539, opened 2022-07-05
Allow user to select language#373

pepermao:
image


src/pages/code-of-conduct-page.tsx
```diff
@@ -20,7 +20,7 @@ const AboutPage: NextPage<{ data: string }> = () => {
 };
 
 export async function getServerSideProps({ query, locale, locales, req }) {
-    locale = parser.pick(locales, req.language) || locale || "en";
+    locale = req.cookies.default_language || parser.pick(locales, req.language) || locale || "pt";
```
thesocialdev, line 23: Probably a good time to create a function `getLocale({ req, locale, locale})` and move the logic elsewhere to avoid the need to touch many files in the future when we change the language logic for some reason.

thesocialdev (changes requested):
This is great. I have one minor change to request, see comment below.

PR #543, opened 2022-07-06
Add Missing Key On Logout Item

caneppelevitor:


src/components/AletheiaMenu.tsx
```diff
@@ -104,6 +104,7 @@ const AletheiaMenu = () => {
             </Menu.Item>
             {hasSession && <Menu.Item
                 data-cy={"testLogout"}
+                key="/login"
```
thesocialdev, line 107: Maybe it's better to redirect to the home page, what do you think?

thesocialdev (changes requested):

src/components/AletheiaMenu.tsx
caneppelevitor: You're right, it's better.

thesocialdev (approved):

Issue #505, opened 2022-06-27
Model Claim for generic media types


thesocialdev:
## Background Information
Consider the claim contains an object that can nest multiple claim types, e. g. a Twitter claim type can contain a claim sentence or a claim image and we infer that from the content object structure.

## What
- Define content schemas for different types of claim

### Paragraph
- [x] create a data-hash for the paragraph as well?
```json
{
    "type": "paragraph",
    "data-hash": "paragraph-data-hash", 
    "props": {
        "id": "sentence-id",
    },
    "content": ["sentences"],
};
```

### Sentence
- [x] change paragraph sequence for paragraph data-hash
```json
{
    "type": "sentence",
    "data-hash": "sentence-data-hash", //`${this.paragraphSequence}${this.sentenceSequence}${sentenceContent}`
    "props": {
        "id": "sentence-id",
    },
    "content": "sentence-content",
};
```

### Image
- [x] create image schema
```json
{
    "type": "image",
    "data-hash": "image hash",
    "props": {
    },
    "content": "aws-s3-url",
};
```

## Open Questions
- Does this mean that we might need to expand the content as a new Collection in MongoDB? Maybe the nesting logic might add overhead to MongoDB queries.
- Does MongoDB support recursive reference of documents?
- What about Claims that are not from a personality? How do we discover them?
    - We can discover claims without personality by topic or enable a new interface for Claims without personality
    - Also, we can leverage a search capability that is better suited for these new changes

PR #550, opened 2022-07-11
Model Claim for generic media types #505

pepermao:
Define content schemas for different types of claim:
- [x] Create Paragraph document in database
- [x] Create a data-hash for the paragraph
- [x] Create Sentence document in database
- [x] change paragraph sequence for paragraph data-hash
- [ ] Image
## - I think the parse.spec test don't work anymore


server/claim/claim.service.ts
```diff
@@ -36,7 +36,8 @@ export class ClaimService {
             query.personality = Types.ObjectId(query.personality)
         }
         const claims = await this.ClaimModel.find(query)
-            .populate("latestRevision")
+            .populate({ path: 'latestRevision', populate: { path: 'content', select: "content" } })
+            .populate("speeches")
```
thesocialdev, line 40: Maybe this is not needed anymore, right?

thesocialdev (changes requested):

thesocialdev (approved):

PR #583, opened 2022-07-20
Fix mobile carousel

ThaisMap:
- Created a custom hook to allow us to use media queries outside of styled components or style props.
- Used the new hook to adjust the object-fit and object-position of the carousel image when the window is smaller than 768px, which is the ant design's breakpoint between small and medium.

400px
image

700px 
image

1440px
image


thesocialdev (approved):
Awesome, I love it!

PR #627, opened 2022-07-30
Router back when login

### LINKED ISSUE #620: Router back when login
thinking in UX, when we finish login we're redirectioned to home page, but is better if the user is redirectioned to last page he visits

pepermao:
image

closes #620


src/components/Personality/PersonalitiesGrid.tsx
```diff
@@ -1,86 +0,0 @@
-import React from "react";
```
thesocialdev, line 1: This file was deleted by mistake, apparently.

thesocialdev (changes requested):

thesocialdev (approved):

PR #633, opened 2022-07-31
Support image as claims

DuardoN:


server/image/image.service.ts
```diff
@@ -4,28 +4,46 @@ import { HistoryService } from "../history/history.service";
...
+        private historyService: HistoryService,
+        private fileManagementService: FileManagementService
     ) {}
 
     async create(image, user) {
-        const newImage = await new this.ImageModel(image).save();
+        return this.fileManagementService
+            .upload(image)
+            .then(async (imageUploaded) => {
+                const imageDataHash = md5(
```
thesocialdev, line 23: We should create a hash from the Image Buffer. Not the URL details.

thesocialdev (changes requested):

package.json
```diff
@@ -69,17 +68,19 @@
         "@nestjs/core": "^8.0.6",
         "@nestjs/mapped-types": "^1.0.1",
         "@nestjs/mongoose": "^8.0.0",
-        "@nestjs/platform-express": "^8.0.6",
+        "@nestjs/platform-express": "9.0.5",
```
thesocialdev, line 71: Don't forget to bump all related nestjs packages versions together.

server/claim/claim.controller.ts
```diff
@@ -192,20 +205,67 @@ export class ClaimController {
         );
     }
 
+    @IsPublic()
+    @Get("no-personality/claim")
```
thesocialdev, line 209: We can remove the `no-personality` prefix here and use just `claim`

PR #654, opened 2022-08-04
RBAC. NestJs Dynamic Role Guard Implementation.

caneppelevitor:
The MVP doesn't work cause the guard only check the endpoint and for the page that we apply this guard uses the same endpoint as another one that we don't want to affect.


server/auth/roles.guard.ts
```diff
@@ -0,0 +1,28 @@
...
+            "roles",
+            context.getHandler()
+        );
+        const request = context.switchToHttp().getRequest();
+
+        if (request?.user) {
+            const { _id } = request.user;
+            const user = await this.userService.getById(_id);
+
+            return roles.includes(user.role);
```
thesocialdev, line 23: We should always get the role from the ory session.

server/users/schemas/user.schema.ts
```diff
@@ -16,6 +16,9 @@ export class User {
     @Prop({ required: true, unique: true })
     oryId: string;
 
+    @Prop({ required: true })
+    role: string;
```
thesocialdev, line 20: We shouldn't store the roles in our DB. Instead, use ory and collect that information form the session.

One of the reasons for that is because this information need to be consistent and being able to decode that from the session means that we also can do it from the frontend clients.

thesocialdev (changes requested):

server/ability/abilities.guard.ts
```diff
@@ -0,0 +1,57 @@
...
+                ForbiddenError.from(ability).throwUnlessCan(
+                    rule.action,
+                    rule.subject
+                )
+            );
+
+            return true;
+        } catch (error) {
+            if (error instanceof ForbiddenError) {
+                throw new ForbiddenException(error.message);
```
thesocialdev, line 53: Let's create an issue to create the Forbidden page like the NotFound (404) page.

thesocialdev (approved):

server/ability/abilities.guard.ts
caneppelevitor: #673 Created.

PR #658, opened 2022-08-05
Redirect to 404 when trying to access a wrong sentence

### LINKED ISSUE #639: Redirect to 404 when trying to access a wrong sentence
### Background information
Right now there's no error if we try to access a wrong sentence hash but we should redirect to the 404 page.

Image

pepermao:
closes #639


thesocialdev (changes requested):
I think we should also change the sitemap.xml and robots.txt for this PR and do not change the config yet because we're still working in the SEO reports.

thesocialdev (approved):

thesocialdev (approved):

PR #664, opened 2022-08-07
Add config to make every page protected by login

### LINKED ISSUE #634: Add config to make every page protected by login
## Background Information
In order to avoid the test environment to be weaponized by bad actors, we need to enforce login for every page.

## How
Make a config variable that overrides `@isPublic` decorator

pepermao:
closes #634

### DIFF
--- a/server/auth/session.guard.ts
+++ b/server/auth/session.guard.ts
@@ -1,10 +1,7 @@
-import {
-    CanActivate,
-    ExecutionContext, Injectable,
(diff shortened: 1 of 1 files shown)


thesocialdev:
I think we should also change the sitemap.xml and robots.txt for this PR and do not change the config yet because we're still working in the SEO reports.

thesocialdev (approved):

PR #667, opened 2022-08-08
Fix kanban query error

pepermao:
- Since we save machine value as an object we need to query by key value
image
- Now we can differentiate if the task is in draft or not

### DIFF
--- a/server/claim-review-task/claim-review-task.controller.ts
+++ b/server/claim-review-task/claim-review-task.controller.ts
@@ -25,7 +25,7 @@ export class ClaimReviewController {
         private claimReviewTaskService: ClaimReviewTaskService,
         private captchaService: CaptchaService,
         private viewService: ViewService
-    ) { }
+    ) {}
 
     @IsPublic()
     @Get("api/claimreviewtask")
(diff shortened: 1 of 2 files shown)


server/claim-review-task/claim-review-task.controller.ts
```diff
@@ -25,15 +25,14 @@ export class ClaimReviewController {
...
-    ) { }
+    ) {}
 
     @IsPublic()
     @Get("api/claimreviewtask")
     public async getByMachineValue(@Query() getTasksDTO: GetTasksDTO) {
         const { page = 0, pageSize = 10, order = 1, value } = getTasksDTO;
         return Promise.all([
             this.claimReviewTaskService.listAll(page, pageSize, order, value),
-            // This should count the number of documents for the original query without pagination
```
thesocialdev, line 36: Any reason for this comment removal?

pepermao: for me, it's clear what this function does, besides that it already exists in other controllers

thesocialdev (approved):

PR #669, opened 2022-08-08
Change claim rating ranking to a numeric scale

### LINKED ISSUE #661: Change claim rating ranking to a numeric scale
### Background information
In order to present claim reviews better on a google search our rating scale should be numeric with a alternative text. 
i.e. 

1. not-fact
2. false
3. misleading
4. unsustainable
5. unverifiable
6. exaggerated
7. arguable
8. true-but
9. true

pepermao:
- reorganize enum classification and start initial value as 1
- change SEO claim rating ranking to a numeric scale
closes #661

### DIFF
--- a/server/claim-review/dto/create-claim-review.dto.ts
+++ b/server/claim-review/dto/create-claim-review.dto.ts
@@ -9,15 +9,15 @@ import { Personality } from "../../personality/schemas/personality.schema";
 import { Claim } from "../../claim/schemas/claim.schema";
 
 export enum ClassificationEnum {
-    "not-fact",
-    "true",
-    "true-but",
-    "arguable",
-    "misleading",
+    "not-fact" = 1,
     "false",
+    "misleading",
     "unsustainable",
-    "exaggerated",
     "unverifiable",
+    "exaggerated",
+    "arguable",
+    "true-but",
+    "true",
 }
 
 export class CreateClaimReview {
--- a/src/machine/enums.ts
+++ b/src/machine/enums.ts
@@ -20,15 +20,15 @@ enum CompoundStates {
 }
 
 enum ClassificationEnum {
-    "not-fact",
-    "true",
-    "true-but",
-    "arguable",
-    "misleading",
+    "not-fact" = 1,
     "false",
+    "misleading",
     "unsustainable",
-    "exaggerated",
     "unverifiable",
+    "exaggerated",
+    "arguable",
+    "true-but",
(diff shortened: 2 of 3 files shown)


src/machine/enums.ts
```diff
@@ -20,15 +20,15 @@ enum CompoundStates {
 }
 
 enum ClassificationEnum {
```
thesocialdev, line 22: Should we use only one Enum to avoid updating it in 2 different places?

src/pages/claim-review.tsx
```diff
@@ -34,9 +42,9 @@ const ClaimPage: NextPage<{ personality, claim, sentence, sitekey, href, claimRe
         claimReviewed: sentence.content,
         reviewRating: {
             "@type": "Rating",
-            ratingValue: review,
-            bestRating: "true",
-            worstRating: "false",
+            ratingValue: ClassificationEnum[review],
+            bestRating: 1,
+            worstRating: 9,
```
thesocialdev, line 47: Isn't 9 equal to `true`?

server/claim-review/dto/create-claim-review.dto.ts
```diff
@@ -9,15 +9,15 @@ import { Personality } from "../../personality/schemas/personality.schema";
 import { Claim } from "../../claim/schemas/claim.schema";
 
 export enum ClassificationEnum {
-    "not-fact",
-    "true",
-    "true-but",
-    "arguable",
-    "misleading",
+    "not-fact" = 1,
```
thesocialdev, line 12: I think we should leave `not-fact` as 0 and leave it out of the rating for ClaimReview schema, since not-fact is not important and is neither better/worst than the others.

server/claim-review/dto/create-claim-review.dto.ts
```diff
@@ -9,15 +9,15 @@ import { Personality } from "../../personality/schemas/personality.schema";
...
     "false",
+    "misleading",
     "unsustainable",
-    "exaggerated",
     "unverifiable",
+    "exaggerated",
+    "arguable",
+    "true-but",
```
thesocialdev, line 19: Let's take this opportunity to rename true to `trustworthy`. We already that for i18n labels, but we should change for enums too.

src/machine/enums.ts
pepermao: i can't do it, since i want to use in server folder and in src folder

PR #687, opened 2022-08-12
Display default flag if language cookie is not set

### LINKED ISSUE #637: Empty select language.
image
When we open the Aletheia site for the first time, displays nothing on the select.

ThaisMap:
- If the default_language cookie is not set, display the brazilian flag
- Remove the background and arrow icon from the select component 
image

Closes #637

### DIFF
--- a/src/components/Header/SelectLanguage.tsx
+++ b/src/components/Header/SelectLanguage.tsx
@@ -1,8 +1,7 @@
-import React, { useEffect, useState } from "react";
 import { Select } from "antd";
-import { GB, BR } from "country-flag-icons/react/3x2";
-import colors from "../../styles/colors";
+import { BR, GB } from "country-flag-icons/react/3x2";
 import Cookies from "js-cookie";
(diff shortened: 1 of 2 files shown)


src/components/Header/SelectLanguage.tsx
```diff
@@ -24,21 +24,21 @@ const SelectLanguage = (props: { defaultLanguage; dataCy }) => {
...
             data-cy={props.dataCy}
         >
-            <Option default value="pt">
-                <BR style={{ width: "22px" }} />
+            <Option default value="pt" data-cy="testLanguagePt">
+                <BR style={{ width: "25px" }} />
             </Option>
-            <Option value="en">
-                <GB title="EN" style={{ width: "22px" }} />
+            <Option value="en" data-cy="testLanguagePt">
```
thesocialdev, line 40: The data-cy here should be different, right?

thesocialdev (changes requested):

thesocialdev (approved):

PR #703, opened 2022-08-17
Provide media query breakpoints with redux

### LINKED ISSUE #698: Provide useMediaQuery data to every component with redux

ThaisMap:
Structure media queries breakpoints and provide them using redux in the variable vw

Closes #698


src/components/Footer/Footer.tsx
```diff
@@ -1,15 +1,21 @@
...
+
+import { useAppSelector } from "../../store/store";
 import colors from "../../styles/colors";
 import AletheiaSocialMediaFooter from "./AletheiaSocialMediaFooter";
 import FooterInfo from "./FooterInfo";
 
 const Footer = () => {
     const { t } = useTranslation();
-    const isMobile = useMediaQuery("(max-width: 767px)");
+    const { isMobile } = useAppSelector((state) => {
```
thesocialdev, line 13: Looking good!

Maybe it will be more semantic if we stop using `isMobile` and start using the breakpoints directly, what do you think? Here's an example:

```
const layout = useAppSelector(state => state.breakpoints)
[...]
<Row
                        style={{
                            marginTop: "10px",
                            width: "100%",
                            textAlign: "center",
                            flexDirection: "column",
                            marginBottom: layout.sm ? "32px" : "0",
                        }}
                    >
[...]
```

thesocialdev: I'm also okay with naming this new variable to `vw` as in video width, since this is related to CSS as long as we document that in the code.

thesocialdev (approved):

PR #740, opened 2022-09-12
Update header layout

### LINKED ISSUE #602: Update header to match design
### Background information
The header component is not matching the design yet:

How it is 

Image

Design
 
Image 

Main changes
- [x] Fixed search bar (desktop only)
- [x] User icon for account related actions (confirm format with designers)
- [x] Confirm with designers about the language picker

ThaisMap:
- the search input is fixed in desktop mode
- the login, logout and my account options are located under the user icon in the header
- the logo is now a svg without any fixed padding and can receive any color 
- the sidebar is now a drawer and clicking away will close it
- updated cypress tests

image
image
image
image

Closes #602


pepermao:
The height of the header is now 70px, I feel that some header items are oversized, like the height of the search input. In the design the height of this header is 54px, what do you think of reducing this header height and consequently decreasing the items in it proportionally?

image

image

pepermao (changes requested):

thesocialdev:
> When we are in the mobile version, the search component is hidden as it used to be, look at this prototype on the figma improvements page
> 
> image

This is not desirable, the mockup shouldn't be considered. The reason is because the header is too big and will take too much space from the screen.

pepermao:
The last question, @thesocialdev, @ThaisMap should make the header fixed now that it takes up a small part of the screen?

PR #746, opened 2022-09-15
Refactor dynamic form

### LINKED ISSUE #546: Transition between ClaimReviewView to SentenceReportView.
When we finish a claim review the page don't render the state of the machine again without a refresh.

Refactor dynamic form, creating machine in `claim-review.tsx` instead `dynamicForm.tsx` and passing machine to dynamicForm as parameter

### LINKED ISSUE #645: Match header of ClaimReview with SentenceReportView.
Should be : 
image

how it is: 
image

ThaisMap:
Refactor the dynamic form component so that it no longer  creates and manages the state machine
- The state machine is created and provided to the claim review view using a react context provider
- The claim review view uses the current state of the provided machine to switch between the Form view and the Sentence Report view
- The logic to select the forms and events was extracted to separated files and changed to use maps
- The is loading state follows the possible events of current state
- All events use the same function to collect the form data and send it to the machine interpreter
- We now show the same header in all states of the review process

Attention point
- The auto save draft was temporarily disabled because it was breaking the user id format saved in the database

Closes #645 
Closes #546


src/components/ClaimReview/form/DynamicForm.tsx
```diff
@@ -328,19 +252,13 @@ const DynamicForm = ({ sentence_hash, personality, claim, sitekey }) => {
...
-                                              "Fact-checking workflow"
-                                          );
-                                      }
-                            }
+                            onClick={() => {
+                                if (
+                                    event === ReviewTaskEvents.goback ||
+                                    event === ReviewTaskEvents.draft
+                                )
+                                    handleSendEvent(event);
```
pepermao, line 260: That way we don't get the track of events other than goback and draft

pepermao (changes requested):

src/machine/getNextEvent.ts
```diff
@@ -0,0 +1,34 @@
+import { ReviewTaskEvents, ReviewTaskStates } from "./enums";
+
+const getNextEvents = (param: ReviewTaskEvents | ReviewTaskStates) => {
+    const eventsMap = {
+        0: [ReviewTaskEvents.assignUser],
```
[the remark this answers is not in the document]
ThaisMap: When going back from the 'assigned' state to the 'unassigned' the param received by the function is 0

src/components/ClaimReview/form/DynamicForm.tsx
ThaisMap: Actually we do get the tracking because the submit event calls the handleSendEvent after the validations

pepermao (approved):

PR #757, opened 2022-10-19
Text feedback for captcha requirement.

### LINKED ISSUE #711: [BUG] Give a feedback for the user that captcha needs to be filled
Maybe through a hover or something like that

caneppelevitor:
Closes #711 
image

image

### DIFF
--- a/src/components/Claim/ClaimCreate.tsx
+++ b/src/components/Claim/ClaimCreate.tsx
@@ -265,6 +265,18 @@ const ClaimCreate = ({
                         sitekey={sitekey}
                         onChange={onChangeCaptcha}
                     />
+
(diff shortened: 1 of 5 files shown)


src/components/Claim/ClaimCreate.tsx
```diff
@@ -265,6 +265,18 @@ const ClaimCreate = ({
                         sitekey={sitekey}
                         onChange={onChangeCaptcha}
                     />
+
+                    {disableSubmit && (
```
thesocialdev, line 269: Should it be checking for recaptcha string like the others?

thesocialdev (approved):

PR #772, opened 2022-10-25
Video Card Creation

### LINKED ISSUE #770: Video CTA card

caneppelevitor:
Closes #770 
image

### DIFF
--- a/public/locales/pt/videoCard.json
+++ b/public/locales/pt/videoCard.json
@@ -1,3 +1,3 @@
 {
-    "content": "AletheiaFact.org é um movimento e plataforma coletivo de verificação de fatos que imagina uma"
+    "content": "AletheiaFact.org é um movimento e plataforma coletiva para verificação de fatos que imagina uma sociedade onde todos podem ter livre acesso e se envolver com informações verdadeiras e confiáveis com autonomia"
 }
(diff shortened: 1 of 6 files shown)


public/locales/pt/videoCard.json
```diff
@@ -0,0 +1,3 @@
+{
+    "content": "AletheiaFact.org é um movimento e plataforma coletivo de verificação de fatos que imagina uma"
```
thesocialdev, line 2: The text is cut

src/components/VideoCTACard.tsx
```diff
@@ -0,0 +1,46 @@
...
+                }}
+            >
+                {t("videoCard:content")}
+            </p>
+            <div
+                style={{
+                    padding: "16px 30px 27px 30px",
+                }}
+            >
+                <iframe
```
thesocialdev, line 33: We should have a video component, since this is also being used in the home page

public/locales/pt/videoCard.json
caneppelevitor: This is the text that we have in figma.

thesocialdev: Full text:
```
AletheiaFact.org is a movement and collective platform for fact-checking that imagines a society where everyone can have free access and engage with truthful and credible information with autonomy
```
See translations in the about page.

thesocialdev (approved):

PR #811, opened 2022-11-18
Unauthorized page for UnauthorizedExption

caneppelevitor:
image

### DIFF
--- a/server/view/view.controller.ts
+++ b/server/view/view.controller.ts
@@ -1,4 +1,12 @@
-import { Controller, Get, Res, Req, Optional, Header } from "@nestjs/common";
+import {
+    Controller,
+    Get,
+    Res,
+    Req,
+    Optional,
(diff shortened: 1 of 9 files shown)


server/view/view.controller.ts
```diff
@@ -104,4 +112,25 @@ export class ViewController {
             .getNextServer()
             .render(req, res, "/404-page", Object.assign(parsedUrl.query));
     }
+
+    @IsPublic()
+    @Get("unauthorized")
+    @Header("Cache-Control", "max-age=86400")
```
thesocialdev, line 118: If you want to use cached response, you shouldn't use variables in the Unauthorised message, like rolê and previous url.

thesocialdev (changes requested):

thesocialdev (approved):

PR #812, opened 2022-11-23
Fix code smells and duplications

ThaisMap:
Fix most of the code smells of these types:
- unnused imports
- unnused variables
- commented out code
- Nested ternary operations
- Unnecessary catch clauses
- Unnecessary useEfects
- use for of with iterable
- assignment within return
- some dependency arrays
- assign object to a variable before exporting as default 

Also removed the duplicated code on schemas


server/parser/parser.service.ts
```diff
@@ -4,37 +4,37 @@ import { ParagraphService } from "../paragraph/paragraph.service";
...
 
-// TODO: regex for future rules
-// const alphabets = /([A-Za-z])/g;
-const prefixes = /(Mr|St|Mrs|Ms|Dr)[.]/g;
-const phdRegex = /Ph\.D\./g;
-// const suffixes = "(Inc|Ltd|Jr|Sr|Co)";
-// const starters = /(Mr|Mrs|Ms|Dr|He\s|She\s|It\s|They\s|Their\s|Our\s|We\s|But\s|However\s|That\s|This\s|Wherever)/g;
-// const acronyms = /([A-Z][.][A-Z][.](?:[A-Z][.])?)/g;
-// const websites = /[.](com|net|org|io|gov)/g;
+/* TODO: regex for future rules
```
thesocialdev, line 11: We can remove this, compromise already cover most of it.

Another solution for code quality is to add test cases for each of these class of problems.

PR #817, opened 2022-11-30
Stop users from uploading the same file multiple times

### LINKED ISSUE #815: Remove the possibility of uploading duplicate images
## Background information
We use the content data hash to manage the claim review tasks, and if users upload the same image the data hash generated will be the same. So we shouldn't allow users to upload the same image multiple times. 

## Suggestion
This can be achieved by checking the data_hash before uploading and returning a [status 303](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303) to indicate that the image is already in the database.

ThaisMap:
Send a status 303 when trying to upload an image that already exists in the platform. 

image

Closes #815

### DIFF
--- a/server/image/image.controller.ts
+++ b/server/image/image.controller.ts
@@ -1,20 +1,35 @@
-import { Controller, Post, UseInterceptors } from "@nestjs/common";
+import { Controller, HttpStatus, Post, UseInterceptors } from "@nestjs/common";
 import {
-    Req,
+    Res,
     UploadedFiles,
 } from "@nestjs/common/decorators/http/route-params.decorator";
 import { FilesInterceptor } from "@nestjs/platform-express/multer";
 import { FileManagementService } from "../file-management/file-management.service";
+import { ImageService } from "./image.service";
+const md5 = require("md5");
 
 @Controller()
 export class ImageController {
-    constructor(private fileManagementService: FileManagementService) {}
+    constructor(
+        private fileManagementService: FileManagementService,
+        private imageService: ImageService
+    ) {}
 
     @Post("api/image")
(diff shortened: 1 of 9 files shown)


server/image/image.controller.ts
```diff
@@ -1,20 +1,35 @@
...
-    async upload(@UploadedFiles() files: Express.Multer.File[], @Req() req) {
+    async upload(@UploadedFiles() files: Express.Multer.File[], @Res() res) {
         return Promise.all(
             files.map(async (file) => {
+                const imageDataHash = md5(file.buffer);
+                const foundImage = await this.imageService.getByDataHash(
+                    imageDataHash
+                );
+                if (foundImage) {
+                    return res.status(HttpStatus.SEE_OTHER).json({
```
thesocialdev, line 28: This should be sending the target link for the image that already exists so the browser can do the redirect.

thesocialdev (changes requested):
Looking good. We need now to solve the redirect part of the response, since 303 implies a redirect action.

src/components/Claim/CreateClaim/ClaimUploadImage.tsx
```diff
@@ -62,8 +62,12 @@ const ClaimUploadImage = () => {
                         router,
                     });
                 })
-                .catch(() => {
+                .catch((err) => {
                     setIsloading(false);
+                    const seeOtherTarget = err.response.data.target;
+                    if (seeOtherTarget) {
```
thesocialdev, line 68: We should also check if status code is 303, otherwise it's an unknown error.

thesocialdev (approved):

thesocialdev (approved):

PR #819, opened 2022-12-07
Rename sentence hash

ThaisMap:
Creates migrations to rename the sentence_hash to data_hash field in the claimReview, reports and claimReviewTasks schemas to better reflect the possibility of reviewing contents that are not sentences.
Updates the code base where needed to use the data_hash fields.


migrate-mongo-config-example.ts
```diff
@@ -1,9 +1,12 @@
 // In this file you can configure migrate-mongo
-
 const config = {
     mongodb: {
+        // Change (or review) the url to your MongoDB:
         url: "MONGODB_URI",
 
+        // Change this to your database name:
+        databaseName: "Aletheia",
```
thesocialdev, line 8: The database name should be present in the mongodb uri, otherwise we can't take advantage of current CI/CD secrets configurations.

thesocialdev (changes requested):

thesocialdev (approved):

PR #866, opened 2023-04-14
Using correct default prop on Image Schema to match ContentEnum.

### LINKED ISSUE #864: Image claim creation is broken after claim review
https://test.aletheiafact.org/claim/643875aaab9111003098b6cd

caneppelevitor:
Closes #864

### DIFF
--- a/server/claim-revision/schema/claim-revision.schema.ts
+++ b/server/claim-revision/schema/claim-revision.schema.ts
@@ -2,14 +2,9 @@ import { Prop, Schema, SchemaFactory } from "@nestjs/mongoose";
 import * as mongoose from "mongoose";
 import { Claim } from "../../claim/schemas/claim.schema";
 import { Personality } from "../../personality/schemas/personality.schema";
+import { ContentModelEnum } from "types/enums";
 
 export type ClaimRevisionDocument = ClaimRevision & mongoose.Document;
-
-export enum ContentModelEnum {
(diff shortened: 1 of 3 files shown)


server/image/schemas/image.schema.ts
```diff
@@ -6,7 +6,7 @@ export type ImageDocument = Image & mongoose.Document;
 @Schema({ toObject: { virtuals: true }, toJSON: { virtuals: true } })
 export class Image {
     @Prop({
-        default: "image",
+        default: "Image",
```
thesocialdev, line 9: What do you think about using a Typescript enum?

thesocialdev (changes requested):

server/claim-revision/schema/claim-revision.schema.ts
```diff
@@ -2,15 +2,10 @@ import { Prop, Schema, SchemaFactory } from "@nestjs/mongoose";
 import * as mongoose from "mongoose";
 import { Claim } from "../../claim/schemas/claim.schema";
 import { Personality } from "../../personality/schemas/personality.schema";
+import { ContentModelEnum } from "types/enums";
```
thesocialdev, line 5: **issue (blocking):** this is breaking the build, I believe you should use a relative path.

thesocialdev (approved):

PR #868, opened 2023-04-18
Removed view button on claimCard when it is a debate.

caneppelevitor:


src/components/Claim/ClaimCard.tsx
```diff
@@ -170,24 +171,26 @@ const ClaimCard = ({ personality, claim, collapsed = true }) => {
...
-                        <span
-                            style={{
-                                color: colors.white,
-                                fontSize: 16,
-                                fontWeight: 400,
-                                margin: 0,
-                                padding: 0,
-                                lineHeight: "24px",
-                            }}
+                    {isDebate && (
```
thesocialdev, line 174: **question:** should it be `!isDebate`?

thesocialdev: Also, isn't that going to break the card in the personality page?

thesocialdev (approved):

PR #912, opened 2023-05-22
Form inputSearch displaying user label instead of userId

### LINKED ISSUE #897: BUG: User Id is being showed in cross-checking
## Background Information

For some reason it is returning the user ID when we return states that has the select user component

image

caneppelevitor:
Closes #897

### DIFF
--- a/server/users/users.controller.ts
+++ b/server/users/users.controller.ts
@@ -167,7 +167,6 @@ export class UsersController {
         return this.usersService.findAll(getUsers);
     }
 
-    @IsPublic()
     @Get("api/user/:id")
     @Header("Cache-Control", "max-age=60, must-revalidate")
     public async getUser(@Param("id") userId) {
--- a/server/users/users.controller.ts
+++ b/server/users/users.controller.ts
@@ -166,4 +166,12 @@ export class UsersController {
     public async getAll(@Query() getUsers) {
         return this.usersService.findAll(getUsers);
     }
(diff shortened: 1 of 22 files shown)


server/users/users.controller.ts
```diff
@@ -166,4 +166,12 @@ export class UsersController {
     public async getAll(@Query() getUsers) {
         return this.usersService.findAll(getUsers);
     }
+
+    @IsPublic()
```
thesocialdev, line 170: That's a privacy breach. We shouldn't allow this endpoint to be public. Are there any reasons for that?

src/components/Form/Autocomplete.tsx
```diff
@@ -18,12 +21,49 @@ const Autocomplete = ({
...
     dataCy,
     dataLoader,
+    type,
     mode = "multiple",
     style = {},
     value = [],
     preloadedOptions = [],
     preloadedTopics = [],
 }: AutocompleteProps) => {
+    const [treatedValue, setTreatedValue] = useState([]);
```
thesocialdev, line 31: **issue(blocking):** this logic should be in a input wrapper just for user. I suggest you create a new input called `UserInput` and add this logic there, and touch the autocomplete component as little as possible.

thesocialdev (changes requested):

thesocialdev (approved):

PR #915, opened 2023-05-22
Added Placeholder For Empty Badges on Profile.

### LINKED ISSUE #890: Show placeholder for empty badges in "My achievements" in profile page
If the user doesn't have any achievements, we could show more information, not just a title

image

caneppelevitor:
Closes #890 
image
image

### DIFF
--- a/public/locales/en/profile.json
+++ b/public/locales/en/profile.json
@@ -1,6 +1,7 @@
 {
     "pageTitle": "My Account",
     "badgesTitle": "My Accomplishments",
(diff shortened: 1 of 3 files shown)


caneppelevitor:
@thesocialdev About this task it would be cool if we design a way to connect the users with future events that will give badges.

thesocialdev:
> @thesocialdev About this task it would be cool if we design a way to connect the users with future events that will give badges.

That's a great idea! Make sure to bring this to the design team!

thesocialdev (approved):

PR #922, opened 2023-05-24
Migrate CI/CD to AWS

thesocialdev:


config.yml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
```
thesocialdev, line 1: issue (blocking): `ory_access_token` is a secret leak, we need to change how we pull this into the config map. Can every one of these be pulled from environment variables?

PR #923, opened 2023-05-29
Search Page With Filters:

caneppelevitor:
- Manage filter query parameter and adapted functions to make it possible search with and without filter parameters
- Adapted Sentence findAll to be able to search by filter without necessary a searchText
- Create a Advanced Search Component to manage filters on the search
- Made topic tags clickable redirecting user to search page with the tag as the main filter

### DIFF
--- a/server/claim/types/sentence/sentence.service.ts
+++ b/server/claim/types/sentence/sentence.service.ts
@@ -4,6 +4,14 @@ import { SentenceDocument, Sentence } from "./schemas/sentence.schema";
 import { InjectModel } from "@nestjs/mongoose";
 import { ReportService } from "../../../report/report.service";
 
+interface FindAllOptionsFilters {
+    searchText: string;
+    pageSize: number;
+    language?: string;
+    skipedDocuments?: number;
+    filter?: string | string[];
+}
+
 @Injectable()
 export class SentenceService {
     constructor(
@@ -41,8 +49,51 @@ export class SentenceService {
         return this.SentenceModel.updateOne({ _id: sentence._id }, newSentence);
     }
 
-    async findAll(searchText, pageSize, skipedDocuments, filter) {
-        let pipeline: object[] = [
+    async findAll({
+        searchText,
+        pageSize,
+        skipedDocuments,
+        filter,
+    }: FindAllOptionsFilters) {
+        let pipeline: object[] = [];
+
+        if (searchText) {
+            pipeline.push({
+                $search: {
+                    index: "sentence_fields",
+                    text: {
+                        query: searchText,
+                        path: "content",
+                        fuzzy: {
+                            maxEdits: 2,
+                        },
+                    },
+                },
+            });
+        } else {
+            pipeline.push({
+                $search: {
+                    index: "sentence_fields",
+                    text: {
+                        query: filter,
+                        path: "topics",
+                        fuzzy: {
+                            maxEdits: 2,
+                        },
+                    },
+                },
+            });
+        }
+
+        if (filter) {
+            pipeline.push({
+                $match: {
+                    topics: { $in: Array.isArray(filter) ? filter : [filter] },
+                },
+            });
+        }
+
+        pipeline.push(
             {
                 $lookup: {
                     from: "paragraphs",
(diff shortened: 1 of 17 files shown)


src/api/topicsApi.ts
```diff
@@ -1,19 +1,25 @@
...
 import { message } from "antd";
+import { ActionTypes } from "../store/types";
 
 const request = axios.create({
     withCredentials: true,
     baseURL: `/api/topics`,
 });
 
-const getTopics = (topicName, t) => {
+const getTopics = (dispatch, topicName, t) => {
```
thesocialdev, line 10: suggestions: if you're going to pass optional parameters, pass an object as parameter and do the following:
```
const getTopics = ({ dispatch, topicName, t }) => {
```

thesocialdev: Also, please create a type interface for this function

server/search/search.controller.ts
```diff
@@ -125,7 +135,12 @@ export class SearchController {
                     req.language,
                     null
                 ),
-                this.sentenceService.findAll(searchText, parsedPageSize, null),
+                this.sentenceService.findAll(
+                    searchText,
+                    parsedPageSize,
+                    null,
```
thesocialdev, line 141: suggestion: make it optional in the function statement and just omit the parameter when calling the function

server/claim/types/sentence/sentence.service.ts
```diff
@@ -123,7 +111,45 @@ export class SentenceService {
...
+                        path: "topics",
+                        fuzzy: {
+                            maxEdits: 2,
+                        },
+                    },
+                },
+            });
+        }
+
+        if (filter) {
```
thesocialdev, line 144: question: what happens if searchText is provided without filter? Also, does the order matter?

server/claim/types/sentence/sentence.service.ts
```diff
@@ -123,7 +111,45 @@ export class SentenceService {
                     },
                 },
             },
-        ]);
+        ];
+
+        if (searchText) {
+            pipeline.splice(0, 0, {
```
thesocialdev, line 117: suggestion: The splice logic is a bit confusing to read, I suggest that we just push them in the order that matters. It seems that you're using splice just to add the pipelines in the right order, maybe we should just push them in the right order starting from an empty array, what do you think?

server/search/search.controller.ts
```diff
@@ -36,35 +36,44 @@ export class SearchController {
...
-                        searchText,
-                        processedPageSize,
-                        skipedDocuments
-                    ),
-                    this.claimRevisionService.findAll(
-                        searchText,
-                        processedPageSize,
-                        skipedDocuments
-                    ),
+                    searchText
```
thesocialdev, line 52: suggestion: convert each of these promises into a separate function and use async/await

src/components/ClaimReview/TopicInput.tsx
```diff
@@ -27,7 +27,7 @@ const TopicInput = ({ contentModel, data_hash, topics }) => {
     const [duplicatedErrorMessage, setDuplicatedErrorMessage] = useState("");
 
     const fetchTopicList = async (topic) => {
-        const topicSearchResults = await topicApi.getTopics(topic, t);
+        const topicSearchResults = await topicApi.getTopics(null, topic, t);
```
thesocialdev, line 30: See my comment in the `topicApi` file

src/components/Search/AdvancedSearch.tsx
```diff
@@ -0,0 +1,33 @@
...
+            id="tags-outlined"
+            defaultValue={
+                Array.isArray(defaltValue) ? defaltValue : [defaltValue]
+            }
+            options={options.map((option) => option.name)}
+            onChange={(event, newValue) => handleFilter(newValue)}
+            renderInput={(params) => (
+                <TextField
+                    {...params}
+                    style={{ backgroundColor: "#f8f9fa" }}
```
thesocialdev, line 23: issue: get the color from the colors utils/library object

src/components/Search/SearchPageView.tsx
```diff
@@ -133,21 +168,38 @@ function SearchPageView({ searchText }) {
...
-                                <SearchOutlined />
+                                <SearchIcon />
                             </Button>
                         </div>
+                        <AdvancedSearch
+                            defaltValue={filtersUsed}
+                            onSearch={fetchTopicList}
+                            options={autoCompleteTopicsResults}
+                            handleFilter={handleFilter}
+                            t={t}
```
thesocialdev, line 201: issue: do not pass `t` as a prop, use the hook inside `AdvancedSearch` instead.

thesocialdev (changes requested):

thesocialdev (changes requested):
Looks great! I have a few comments to improve the code readability and a couple of issues.

server/claim/types/sentence/sentence.service.ts
caneppelevitor: Is gonna bring a common search with all results based on the searchText. And about the order of what specific, the splice on the pipe line or about the searchText and filter ?

server/claim/types/sentence/sentence.service.ts
caneppelevitor: I didn't have this ideia, thanks for the suggestion.

thesocialdev (approved):

PR #925, opened 2023-06-09
Bug Fixing and Correction of Sources Title

caneppelevitor:


src/components/SentenceReport/SentenceReportContent.tsx
```diff
@@ -55,10 +55,6 @@ const SentenceReportContent = ({ context }) => {
                 </Col>
             )}
             <Col span={24}>
-                <Paragraph className="title">
```
thesocialdev, line 58: question: I don't see an issue so I didn't understand the reason of this to be removed, can you elaborate?

caneppelevitor: The problem was, when I changed the SourcesList component I added the tittle on the component and removed from ClaimViewPage, but this component was being used on SentenceReportContent causing the duplication of the titles: 
image

ps: But thinking again I gonna change that cause I think its better for the tittle to be outside of the component. What do you think ?

thesocialdev: Agreed.

thesocialdev (approved):

PR #952, opened 2023-08-12
Create collaborative editor web socket service runner

pepermao:


server/websocket.ts
```diff
@@ -0,0 +1,22 @@
+import { NestExpressApplication } from "@nestjs/platform-express";
+import { NestFactory } from "@nestjs/core";
+import { WebsocketModule } from "./yjs-websocket/websocket.module";
+import { WsAdapter } from "@nestjs/platform-ws";
+const cookieParser = require("cookie-parser");
+
+const initApp = async (options) => {
+    const { config } = options;
```
thesocialdev, line 8: Get port and other options from the config

server/websocket.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    const app = await NestFactory.create<NestExpressApplication>(
+        WebsocketModule
+    );
+    app.enableCors();
+    app.use(cookieParser());
+
+    app.useWebSocketAdapter(new WsAdapter(app));
+    await app.listen(config.port);
+    console.log(`Application is running on: ${await app.getUrl()}`);
```
thesocialdev, line 18: Use service-runner logger

src/components/Collaborative/utils/createWebsocketConnection.ts
```diff
@@ -1,7 +1,13 @@
-import { WebsocketProvider } from "y-websocket";
 import * as Y from "yjs";
 
+import { WebsocketProvider } from "y-websocket";
+
 export function createWebsocketConnection(hash: string) {
     const ydoc = new Y.Doc();
-    return new WebsocketProvider("ws://localhost:1234", hash, ydoc);
+
+    // @TODO get this from the config yaml through SSR
```
thesocialdev, line 8: issue(blocking): Complete this todo

src/tsconfig.json
```diff
@@ -1,18 +1,19 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2021",
```
thesocialdev, line 3: chore: maybe we we don't need that anymore now that we migrated the code to node 18 and yarn 3. Let's try to remove this.

thesocialdev (changes requested):
A couple more comments:
- We should stop commiting doc files
- Let's make sure we can have the proper config for the websocket server in place but it's better to have a separate MR to deploy this.

src/tsconfig.json
thesocialdev: Not harmful and it's an improvement. Let's do it.

thesocialdev (approved):

PR #994, opened 2023-09-13
Show Users 2fa Status on Admin Page

### LINKED ISSUE #968: Show in the user table if the user enabled 2fa

caneppelevitor:
Page to check users Authenticator Assurance Level: 
image

AdminPage:
image
Closes #968


server/view/view.controller.ts
```diff
@@ -118,6 +118,22 @@ export class ViewController {
             .render(req, res, "/404-page", Object.assign(parsedUrl.query));
     }
 
+    @IsPublic()
```
thesocialdev, line 121: question: should this be public? The user needs to be logged in so maybe we should enforce that as well.

src/components/TotpCheckPage.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useTranslation } from "next-i18next";
+import { userBeingEdited } from "../atoms/userEdit";
+
+const AalCheckPage = () => {
+    const { t } = useTranslation();
+
+    const [aal] = useAtom(currentAuthentication);
+    const [userId] = useAtom(currentUserId);
+    const [currentUser] = useAtom(userBeingEdited);
+    console.log(currentUser, "kkkkkk");
```
thesocialdev, line 14: chore: remove console.log

thesocialdev (approved):

PR #997, opened 2023-09-14
Depriving Regular Users of Acess The Platform When It is Private

### LINKED ISSUE #966: When the platform is configured as a private one, do not allow regular uses to access the content

caneppelevitor:
Closes #966

### DIFF
--- a/server/auth/session.guard.ts
+++ b/server/auth/session.guard.ts
@@ -47,17 +47,24 @@ export class SessionGuard implements CanActivate {
                     this.configService.get<string>("override_public_routes");
 
                 if (overridePublicRoutes) {
-                    return this.checkRole(request, response, isPublic);
-                } else return true;
+                    return this.checkAndRedirect(
+                        request,
+                        response,
+                        isPublic,
+                        "/unauthorized"
+                    );
+                } else {
+                    return true;
+                }
             }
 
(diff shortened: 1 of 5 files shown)


server/auth/session.guard.ts
```diff
@@ -42,16 +42,39 @@ export class SessionGuard implements CanActivate {
...
             }
 
             return this.next(request, response, isPublic);
         } catch (e) {
-            // TODO: logging
             return this.next(request, response, isPublic);
         }
     }
 
+    checkRole(request, response, isPublic) {
```
thesocialdev, line 60: suggestion: most of the logic is also used in the `next()` function, maybe you should consolidate them into a single function.

thesocialdev (changes requested):

server/auth/session.guard.ts
```diff
@@ -42,33 +42,49 @@ export class SessionGuard implements CanActivate {
                     role: session?.identity?.traits?.role,
                     status: session?.identity.state,
                 };
-                return true;
+                const overridePublicRoutes =
+                    session?.identity?.traits?.role === "regular" &&
```
pepermao, line 46: suggestion: Use `Roles` enum

thesocialdev (approved):

PR #1010, opened 2023-09-15
MVP Notification

caneppelevitor:
## Icon when it has notifications
image
## Notification Menu
image
## Unread notifications
image


src/machines/reviewTask/reviewTaskMachine.ts
```diff
@@ -168,6 +170,14 @@ export const transitionHandler = (state) => {
             })
             .finally(() => resetIsLoading());
     }
+
+    sendReviewNotifications(
```
thesocialdev, line 174: suggestion (non-blocking): if you do this in the backend it's more secure and you would still have access to the needed information. This would also reduce the amount of responsibility of this component.

thesocialdev (approved):

server/notifications/novu.provider.ts
```diff
@@ -0,0 +1,16 @@
...
+import { ConfigService } from "@nestjs/config";
+import { Novu } from "@novu/node";
+
+const NOVU_PROVIDER_TOKEN = "NOVU_PROVIDER_TOKEN";
+
+export const NovuProvider: Provider = {
+    provide: NOVU_PROVIDER_TOKEN,
+    inject: [ConfigService],
+    useFactory: () => {
+        const NOVU_API_KEY = process.env.NOVU_API_KEY;
```
thesocialdev, line 11: issue: this should be pulled from the config.yaml file

thesocialdev (changes requested):

migrate-mongo-config-example.ts
```diff
@@ -20,6 +20,7 @@ const config = {
 
     // The mongodb collection where the applied changes are stored. Only edit this when really necessary.
     changelogCollectionName: "changelog",
+    novu_api_key: "NOVU_API_KEY",
```
thesocialdev, line 23: question: do we really need this?

server/app.module.ts
```diff
@@ -112,6 +113,9 @@ export class AppModule implements NestModule {
                 })
             );
         }
+        if (options.config.novu) {
```
thesocialdev, line 116: chore: let's create a task to re-evaluate this approach. If the module is not registered I believe the services depending on it will also crash, we need something more decoupled, like a message system or something like that.

thesocialdev (approved):

migrate-mongo-config-example.ts
pepermao: We need it to be able to make the migration since we need the API key to have access to Novus's API

PR #1021, opened 2023-09-21
Start workflow implementation inside of the collaborative editor

pepermao:
- [x] Concentrate all fields in the same state


thesocialdev:
Cypress tests will but the main part is being tested in unit tests. Let's fix the cypress tests in a different PR.

PR #1066, opened 2023-10-21
MVP Namespace Implementation

pepermao:
- Create namespace schema and module
- Add optional namespace paremeter
- filter claims by namespace


src/components/Personality/PersonalityCard.tsx
```diff
@@ -50,7 +52,15 @@ const PersonalityCard = ({
     const [state] = useAtom(createClaimMachineAtom);
     const { claimData } = state.context;
     const { personalities } = claimData;
+    const [nameSpace] = useAtom(currentNameSpace);
     const { vw } = useAppSelector((state) => state);
+    console.log("hrefBase: ", hrefBase);
```
thesocialdev, line 57: chore: remove console.log

server/auth/name-space/name-space.service.ts
```diff
@@ -0,0 +1,123 @@
...
+        );
+        await this.findNameSpaceUsersAndDelete(
+            nameSpace.slug,
+            newNameSpace.users,
+            nameSpace.users
+        );
+
+        return await this.NameSpaceModel.updateOne(
+            { _id: nameSpace._id },
+            newNameSpace
```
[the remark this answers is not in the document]
thesocialdev: @pepermao I think this is a good point, we need to sanitise this before doing do update, otherwise it creates a vector to update arbitrary namespaces.

thesocialdev (changes requested):

server/claim/claim.controller.ts
```diff
@@ -209,11 +230,11 @@ export class ClaimController {
     }
 
     @IsPublic()
-    @Get("api/claim/:id")
+    @Get("api/:namespace?/claim/:id")
```
thesocialdev, line 233: question: won't that create a conflict and allow the following endpoint? `:namespace?/api/:namespace?/claim/:id`. You can find that in the mapped endpoints in the startup log.

server/util.ts
```diff
@@ -1,6 +1,7 @@
 import { Injectable } from "@nestjs/common";
 import { Roles } from "./auth/ability/ability.factory";
 import { randomBytes } from "crypto";
+import { NameSpaceEnum } from "auth/name-space/schemas/name-space.schema";
```
thesocialdev, line 4: issue: use a relative url

thesocialdev (changes requested):

PR #1084, opened 2023-11-07
Create Collaborative Editor Cypress Test

pepermao:


config.test.ci.yaml
```diff
@@ -22,6 +22,7 @@ services:
       port: 3000
```
thesocialdev, line 22: issue: This config is not creating a websocket server

cypress.config.ts
```diff
@@ -5,7 +5,7 @@ export default defineConfig({
     video: false,
     screenshotOnRunFailure: false,
     pageLoadTimeout: 480000,
-    defaultCommandTimeout: 30000,
+    defaultCommandTimeout: 65000,
```
thesocialdev, line 8: @pepermao even after doubling the timeout, it's not working so I think there's something wrong with the test itself. Are you sure the review test spec is working fine?

thesocialdev: I think it might have to do with the super-admin role. When running locally it can't even create a claim, the button doesn't appear which I assume means the user is being considered regular.

thesocialdev (approved):

PR #1108, opened 2023-11-19
Reduce ory requests

### LINKED ISSUE #1098: Ory: Stop using the integration package in production
## Background Information
Because we have deployed to production using custom domain, we don't need to have a middleware to integrate Ory and the application, this will improve our calls to the ory endpoints since it won't need to pass through our infrastructure.

## How
- Follow Ory's guide https://www.ory.sh/docs/getting-started/integrate-auth/nextjs#go-to-production.
- We should still use the integration for development and the testing environment.

### LINKED ISSUE #1099: Too many requests to the /whoami endpoint
## Background Information
We have identified that the application is doing to many requests to the whoami page and this can be a scaling issue. Here are some tasks to look at

## What
- [x] #1100
- [x] #1101

### LINKED ISSUE #1101: Ory: Reduce requests during startup time to 1

pepermao:
- Stop using ory/integration in production

closes #1101 
closes #1099
closes #1098


.github/workflows/nodejs.yml
```diff
@@ -87,6 +88,9 @@ jobs:
        SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
   cypress-run:
     needs: build
+    env:
```
thesocialdev, line 91: as discussed synchronously, this won't work in cypress due to CSRF token errors

deployment/app.yml
```diff
@@ -90,6 +90,10 @@ spec:
               value: NEW_RELIC_LICENSE_KEY_STUB
             - name: NEW_RELIC_APP_NAME
               value: NEW_RELIC_APP_NAME_STUB
+            - name: ENVIRONMENT
```
thesocialdev, line 93: issue: now that I'm looking at the full file, I think we will have a problem because ENVIRONMENT is being replaced at the top. If the regex is applied globally, this will mess up with the ENVIRONMENT_STUB

thesocialdev (approved):

PR #1130, opened 2024-03-28
Remove service-runner dependency and refactor code accordingly.

caneppelevitor:


server/websocket.ts
```diff
@@ -2,29 +2,32 @@ import { NestExpressApplication } from "@nestjs/platform-express";
...
-import Logger from "./logger";
+import { Logger } from "@nestjs/common";
+const fs = require("fs");
+const yaml = require("js-yaml");
 
-const initApp = async (options) => {
-    const { config } = options;
+async function initApp() {
+    const defaultConfigFilePath = "config.yaml";
+    const doc = yaml.load(fs.readFileSync(defaultConfigFilePath, "utf8"));
```
thesocialdev, line 11: issue (blocking): use the new `loadConfig` here too

server/seed.ts
```diff
@@ -1,21 +1,23 @@
...
 import { EmailService } from "./email/email.service";
 import { ConfigService } from "@nestjs/config";
 import { UsersService } from "./users/users.service";
 import { UtilService } from "./util";
+import loadConfig from "./configLoader";
 
-const initApp = async (options) => {
-    options.logger.log("info", `Loading AppModule`);
+async function initApp() {
+    const defaultConfigFilePath = "config.seed.yaml";
```
thesocialdev, line 12: suggestion: what do you think about moving `defaultConfigFilePath` to the loadConfig function?

server/main.ts
```diff
@@ -1,27 +1,36 @@
...
+import { ValidationPipe, Logger } from "@nestjs/common";
 import { SwaggerModule, DocumentBuilder } from "@nestjs/swagger";
-require("newrelic");
+import loadConfig from "./configLoader";
+import * as dotenv from "dotenv";
 const cookieParser = require("cookie-parser");
 const mongoose = require("mongoose");
+dotenv.config();
+
+const isLocal = process.env.ENV === "Local";
```
thesocialdev, line 13: suggestion: we are use lower case to the other `ENV` variables, we should do it here too.

thesocialdev: Also, add this to the env-example file

server/main.ts
```diff
@@ -1,27 +1,36 @@
...
+if (!isLocal) {
+    require("newrelic");
+}
+
+async function initApp() {
+    const defaultConfigFilePath = "config.yaml";
```
thesocialdev, line 19: Same as below

server/main.ts
```diff
@@ -1,27 +1,36 @@
...
 import { NestExpressApplication } from "@nestjs/platform-express";
-import { ValidationPipe } from "@nestjs/common";
+import { ValidationPipe, Logger } from "@nestjs/common";
 import { SwaggerModule, DocumentBuilder } from "@nestjs/swagger";
-require("newrelic");
+import loadConfig from "./configLoader";
+import * as dotenv from "dotenv";
 const cookieParser = require("cookie-parser");
 const mongoose = require("mongoose");
+dotenv.config();
```
thesocialdev, line 11: thought (non-blocking): I believe there's a nodejs version that implements what dotenv does and this can be killed. We should find out which one and create an issue for that.

server/configLoader.ts
```diff
@@ -0,0 +1,36 @@
+const fs = require("fs");
+const yaml = require("js-yaml");
+
+function _replaceEnvVars(config) {
```
thesocialdev, line 4: suggestion: add a comment that this was extracted from service-runner codebase

thesocialdev (changes requested):

server/main.ts
caneppelevitor: #1132

thesocialdev (approved):

PR #1150, opened 2024-04-19
ClaimReview e2e Tests

caneppelevitor:


server/claim-review/claim-review.service.ts
```diff
@@ -86,6 +86,7 @@ export class ClaimReviewService {
             { $limit: parseInt(pageSize) }
         );
 
+        // error when uses aggregation
```
thesocialdev, line 89: question: the CI didn't file, should it?

caneppelevitor: Maybe I didn't expressed my self well, it doesn't return an error. The problem is that the aggregation return empty  when the test is ran.

package.json
```diff
@@ -127,6 +129,7 @@
     "migrate-mongo-ts": "^1.1.4",
     "mongoose": "^  5.13.15",
     "mongoose-softdelete-typescript": "^0.0.3",
+    "mongodb-memory-server": "^8.15.1",
```
thesocialdev, line 132: We shouldn't do this. This will download binaries we will never use to the docker image.

thesocialdev (changes requested):

package.json
thesocialdev: We decided to follow-up on a separate MR

thesocialdev (approved):

server/claim-review/claim-review.service.ts
```diff
@@ -36,89 +35,69 @@ export class ClaimReviewService {
...
-            lookupClaims(TargetModel.ClaimReview),
-            { $unwind: "$claim" },
-            lookupClaimRevisions(TargetModel.ClaimReview),
-            { $unwind: "$claim.latestRevision" },
-            {
-                $match: {
-                    "claim.isHidden": query.isHidden,
-                    "claim.isDeleted": false,
-                    "claim.nameSpace":
+        const pipeline = this.ClaimReviewModel.find(query)
```
thesocialdev, line 38: follow-up: we need to add namespace to claim review to avoid bad populate results

PR #1154, opened 2024-04-20
Fixed claim review header responsive card

pepermao:
## Before

image
image

## After

image
image


src/components/ClaimReview/ClaimReviewView.tsx
```diff
@@ -42,6 +49,16 @@ const ClaimReviewView = (props: ClaimReviewViewProps) => {
     const { personality, claim, content, hideDescriptions } = props;
     const isContentImage = claim.contentModel === ContentModelEnum.Image;
 
+    const componentStyle = {
```
thesocialdev, line 52: thought: this is going against the atomic design principles and creating a dependency injection pattern, we should rethink the components in the future to avoid this.

thesocialdev (approved):

PR #1160, opened 2024-04-23
Hotfix: Test environment bugs

pepermao:


src/styles/app.css
```diff
@@ -137,7 +137,7 @@ h6 {
     outline: none;
 }
 
-.MuiPopper-root {
+[data-cy="testFloatingLinkToolbar"] {
```
thesocialdev, line 140: suggestion (non-blocking): we shouldn't identify the element in CSS with the cypress identificator.

thesocialdev (approved):

PR #1187, opened 2024-05-06
Aletheia Chat bot Copilot

pepermao:
Create chat bot user interface to create an interactivity between the user and the artificial intelligence during the fact-checking process

image

## Screen medium size
image
image

## Screen small size
image


server/copilot/copilot-chat.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import { Body, Controller, Post, Req } from "@nestjs/common";
+import { CopilotChatService } from "./copilot-chat.service";
+import { ContextAwareMessagesDto } from "./dtos/context-aware-messages.dto";
+
+@Controller()
+export class CopilotChatController {
+    constructor(private readonly copilotChatService: CopilotChatService) {}
+
+    @Post("api/agent-chat")
```
thesocialdev, line 28: todo: we need to protect this endpoint before the deployment

server/copilot/copilot-chat.service.ts
```diff
@@ -0,0 +1,211 @@
...
+                (message) => this.transformMessage(message)
+            );
+            const tools = [
+                new DynamicStructuredTool(this.getFactCheckingReportTool),
+            ];
+            const currentMessageContent =
+                contextAwareMessagesDto.messages[
+                    contextAwareMessagesDto.messages.length - 1
+                ];
+            console.log("cahed");
```
thesocialdev, line 114: chore: remove console.log

config.example.yaml
```diff
@@ -5,7 +5,7 @@ services:
       cors: '*'
       debug: true
       websocketUrl: ws://localhost:5001
-      agentsUrl: http://localhost:8000
+      automatedFactCheckingAPIUrl: AGENTS_API_URL
```
thesocialdev, line 8: chore: create the deployment for this new config

server/copilot/openAI.constants.ts
```diff
@@ -0,0 +1,16 @@
...
+ * application, helping in maintaining consistency and ease of updates.
+ *
+ * @enum openAI
+ *
+ * @member GPT_3_5_TURBO_1106 - Identifier for the specific OpenAI model version used.
+ * @member BASIC_CHAT_OPENAI_TEMPERATURE - Controls the randomness in the model's responses,
+ *                                         with a higher value resulting in more random responses.
+ */
+export enum openAI {
+    GPT_3_5_TURBO_1106 = "gpt-3.5-turbo-1106",
```
thesocialdev, line 14: follow-up: it would be better to get these from a configuration in the future

server/copilot/copilot-chat.controller.ts
caneppelevitor: I'll pass the Guard to check the users abilities, but I was thinking about to set a special Rate Limiting for this endpoint, what do you think? Do you have any suggestions ?

thesocialdev: The guard should be enough, we just need to make sure we are not spending LLM credits

thesocialdev (approved):

PR #1195, opened 2024-05-14
Hotfix: Properly Connect Sources Between Editor and HTML

caneppelevitor:


thesocialdev (changes requested):
Add test cases for the problem this is trying to solve

server/editor-parse/editor-parse.service.spec.ts
```diff
@@ -154,5 +154,20 @@ describe("ParserService", () => {
...
         });
+
+        it("Source from Schema to HTML is parsed in the right position", async () => {
+            const schemaHtmlResult = await editorParseService.schema2html(
+                schemaContent
+            );
+            expect(schemaHtmlResult.report).toEqual(schemaHtml.report);
+        });
+
+        it("Editor content is parsed correctly with markUp source in the right position", async () => {
```
thesocialdev, line 165: chore (blocking): also add a test case with possibly duplicated texts that this code is attempting to fix.

lib/editor-parser.ts
```diff
@@ -54,8 +54,10 @@ export class EditorParser {
             return fragmentText;
         }
 
-        if (type === "source" && fragmentText === targetText) {
-            return `<a href='#${fragmentText}' rel='noopener noreferrer nofollow'>${fragmentText}<sup>${sup}</sup></a>`;
+        const treatedFragmentText = this.extractTextFromMarkUp(fragmentText);
```
thesocialdev, line 57: nitpick: use the term parsed instead of treated

lib/editor-parser.ts
```diff
@@ -465,4 +479,9 @@ export class EditorParser {
             ],
         }));
     }
+
+    extractTextFromMarkUp(fragmentText) {
+        const match = fragmentText.match(/{{[^|]+\|([^}]+)}}/);
```
thesocialdev, line 484: suggestion: declare the regex as a global variable instead of declaring it here. The reason is because this regex will be compiled in every execution instead of being compiled only once.

server/editor-parse/editor-parse.service.spec.ts
```diff
@@ -154,5 +154,20 @@ describe("ParserService", () => {
             );
             expect(schemaHtmlResult).toMatchObject(schemaHtml);
         });
+
+        it("Source from Schema to HTML is parsed in the right position", async () => {
```
thesocialdev, line 158: praise: nice!

thesocialdev (changes requested):
These are the last requested changes. @pepermao you should review and approve this and feel free to merge.

lib/editor-parser.ts
```diff
@@ -3,6 +3,8 @@ import { ReviewTaskMachineContextReviewData } from "../server/claim-review-task/
 
 const EditorSchemaArray = ["summary", "report", "verification", "questions"];
 
+const regex = /{{[^|]+\|([^}]+)}}/;
```
pepermao, line 6: Suggestion: Rename the regex to be more descriptive

pepermao (approved):

PR #1208, opened 2024-05-25
Create `Unattributed` claim type

pepermao:
- Create Unattributed module
- Fixed /claim page, now listing all claims
- Created a new event in the claimReviewCreate machine, to enable creation of generative information claims
- Changed image endpoint from `claim/claimId` to `claim/claimSlug`

image
image
image
image


server/claim/claim.service.ts
```diff
@@ -378,7 +388,9 @@ export class ClaimService {
                 );
 
             processedClaim.content =
-                processedClaim.contentModel === ContentModelEnum.Speech
+                processedClaim.contentModel === ContentModelEnum.Speech ||
```
thesocialdev, line 391: suggestion (non-blocking): this is getting increasingly complex, we should move to a function and use the strategy pattern.

src/components/Claim/ClaimCard.tsx
```diff
@@ -79,15 +62,21 @@ const ClaimCard = ({ personality, claim, collapsed = true }) => {
             });
             setClaimContent(textContent.trim());
         };
-        if (isSpeech) {
+        if (isSpeech || isGenerativeInformation) {
```
thesocialdev, line 65: suggestion: transform this into a function, in the future we want a better way of identifying the reason to trigger the `CreateFirstParagraph` function.

thesocialdev (changes requested):
Change `Generative Information` to `Unattributed claim`

thesocialdev (approved):

PR #1215, opened 2024-05-27
Daily Report Sender on the Platform.

caneppelevitor:
image

### DIFF
--- a/server/notifications/notification.controller.ts
+++ b/server/notifications/notification.controller.ts
@@ -17,22 +17,15 @@ export class NotificationController {
         private configService: ConfigService
     ) {}
 
-    @Post("api/topic-subscription/:key/send")
+    @Post("api/notification/topic-subscription/:key/send/:nameSpace")
     @UseGuards(AbilitiesGuard)
     @CheckAbilities(new AdminUserAbility())
-    async sendDailyReport(
-        @Body()
-        body: {
-            key: string;
-        }
-    ) {
-        const dailyReport =
-            await this.summarizationService.generateDailyReport();
-
-        return this.notificationService.sendDailyReviewsEmail(
-            body.key,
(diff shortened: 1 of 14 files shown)


src/api/notificationsApi.ts
```diff
@@ -16,10 +17,11 @@ const sendNotification = (subscriberId, payload) => {
         });
 };
 
-const sendEmail = (key) => {
-    return axios
-        .post(`/api/topics/${key}/send`)
+const sendDailyReportEmail = (key, nameSpace, t) => {
+    return request
+        .post(`/topic-subscription/${key}/send/${nameSpace}`)
```
thesocialdev, line 22: issue: we should keep the api prefix

server/notifications/notification.controller.ts
```diff
@@ -17,22 +17,15 @@ export class NotificationController {
...
-    ) {
-        const dailyReport =
-            await this.summarizationService.generateDailyReport();
-
-        return this.notificationService.sendDailyReviewsEmail(
-            body.key,
-            dailyReport
+    async sendDailyReport(@Param("key") key, @Param("nameSpace") nameSpace) {
+        const dailyReport = await this.summarizationService.generateDailyReport(
+            nameSpace
```
thesocialdev, line 25: question: what if the nameSpace doesn't exist?

src/api/notificationsApi.ts
thesocialdev: never mind, I see now the base URL

server/notifications/notification.controller.ts
caneppelevitor: It will get the Main nameSpace.

thesocialdev (approved):

PR #1224, opened 2024-05-29
Added missing verification for cross-checking case on showForm logic.

caneppelevitor:


src/components/ClaimReview/ClaimReviewForm.tsx
```diff
@@ -61,7 +64,8 @@ const ClaimReviewForm = ({
         isUnassigned ||
```
thesocialdev, line 64: issue(blocking): this is getting too complex, please consider moving this to a function and use the "guard clause pattern"

thesocialdev (changes requested):

src/components/ClaimReview/ClaimReviewForm.tsx
```diff
@@ -57,11 +60,16 @@ const ClaimReviewForm = ({
         })
     );
 
-    const showForm =
-        isUnassigned ||
-        userIsAdmin ||
-        (userIsAssignee && !isReviewing) ||
-        (isReviewing && userIsReviewer);
+    const shouldShowForm = () => {
```
thesocialdev, line 63: Issue: this should go in a useEffect now.

thesocialdev (changes requested):

thesocialdev (approved):

PR #1235, opened 2024-06-01
Bug Fixing on Home Feed and Search Results.

caneppelevitor:


src/components/Sentence/SentenceCard.tsx
```diff
@@ -18,8 +18,12 @@ const StyledComment = styled(Comment)`
 `;
 
 const SentenceCard = ({ sentence }) => {
-    const { content, claim, personality } = sentence;
-    const contentModel = sentence?.claim[0].contentModel;
+    const {
+        content,
+        claim: [claim],
```
thesocialdev, line 23: issue: I don't see the point of converting to array just to get the first element later. You should check if it's an array or not and try to access the contentModel based on that

pepermao (approved):

PR #1261, opened 2024-06-18
Rename claim review task files

pepermao:


migrations/20240618124843-change-review-task-history-target-model-field.ts
```diff
@@ -0,0 +1,19 @@
+import { Db } from "mongodb";
+
```
thesocialdev, line 2: question: will we need to change the table name as well?

pepermao: Yes, we do need. To do that in my local environment I exported the `claimreviewtasks` document, created a new one named `reviewtasks` and imported the data

thesocialdev: Can't we have a migration to rename the collection?

thesocialdev (approved):

PR #1263, opened 2024-06-20
MVP: Aletheia Chat Bot

caneppelevitor:
image


server/chat-bot/chat-bot.controller.ts
```diff
@@ -0,0 +1,16 @@
+import { Controller, Post, Req, Res } from "@nestjs/common";
+import { ChatbotService } from "./chat-bot.service";
+import type { Request, Response } from "express";
+
+@Controller()
+export class ChatbotController {
+    constructor(private readonly chatBotService: ChatbotService) {}
+
+    @Post("api/chat-bot/hook")
```
thesocialdev, line 9: issue: we now need to define how we are going to authenticate this request.

caneppelevitor:
> The code looks great. But what happen when the user sent an image?

Currently, to complete this MVP, we are only allowing text messages. If a user sends a message with a different type, such as an image, the chatbot will provide feedback informing the user that only text messages are accepted.

server/chat-bot/chat-bot.controller.ts
thesocialdev: We can use M2M with ory to authenticate with OAUTH2. This would be a very sensitive endpoint that could be exploited by bad actors, we can't proceed without minimal authentication.

caneppelevitor: Paired with @snowmous3 and updated to middleware with authentication.

thesocialdev (approved):

PR #1277, opened 2024-07-02
HotFix: Refactored MarkUp Sources Logic

caneppelevitor:


thesocialdev:
Why did we change the heuristics of the source identification? What's the reason for the sourceId to not work?

caneppelevitor:
> Why did we change the heuristics of the source identification? What's the reason for the sourceId to not work?

This adjustment was necessary because, when a link is pasted (such as from a document), it does not generate the markup ID we were using as a UUID to identify the text with a source, as expected by Remirror. In these cases, the markup ID did not exist, causing issues in text identification and source tracking. To address this problem, I have implemented a manual approach for enumerating the sources, as previously suggest.

thesocialdev (approved):

PR #1341, opened 2024-08-09
fixed: Inserted return statements in the migrations 

Keikonichi:
Inserted return statements in the migrations to fix the error that was occurring when running yarn migrate.

### DIFF
--- a/server/auth/name-space/name-space.guard.ts
+++ b/server/auth/name-space/name-space.guard.ts
@@ -1,6 +1,7 @@
 import {
     CanActivate,
     ExecutionContext,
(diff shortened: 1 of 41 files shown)


server/auth/name-space/name-space.guard.ts
```diff
@@ -10,6 +11,7 @@ import { Configuration, FrontendApi } from "@ory/client";
 import { NameSpaceService } from "./name-space.service";
 
 @Injectable()
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
pepermao, line 14: Question: @Keikonichi could you explain me why this comment line ?

pepermao (approved):

PR #1371, opened 2024-09-10
Fix: ensure sitemap.xml is accessible in production environment

jeffDeveloperFullStack:
Resolved an issue where the `sitemap.xml` was not being properly generated or accessible in the production environment. This fix ensures that the sitemap is correctly generated and can be accessed by search engines, allowing the site to be indexed by Google.

- Updated the sitemap generation script to run correctly in the production environment.

This change should improve the site's visibility in search engine results.

### DIFF
--- a/server/personality/personality.service.ts
+++ b/server/personality/personality.service.ts
@@ -109,9 +109,15 @@ export class PersonalityService {
             );
         }
 
-        return Promise.all(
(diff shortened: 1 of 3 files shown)


thesocialdev (approved):

thesocialdev:
Seems like a flaky test. Merging anyway.

PR #1373, opened 2024-09-13
Adding a debouncing mechanism when searching wikidata topics

### LINKED ISSUE #1326: Create a debouncing mechanism while searching for wikidata's topics
### Background Information
Debouncing is a technique used to ensure that a function is only executed after a certain period of time has passed since it was last invoked.

### How
Create a debouncing logic similar as the one already in used in either the `useAutoSaveDraft` or `InputSearch` component files

### Where
`TopicDisplay.tsx`
```js
const fetchTopicList = async (
        topic: string
    ): Promise<{ label: string; value: string }[]> => {
        const topicSearchResults = await TopicsApi.getTopics({
            topicName: topic,
            t: t,
            dispatch: dispatch,
        });

        return (
            topicSearchResults?.map((topic) => ({
                label: topic.name,
                value: topic.wikidata,
            })) || []
        );
    };
```

LuizFNJ:
Adding a debouncing mechanism ensures that the function is only executed after a period of inactivity. This prevents the function from running with each character.
Example:
Captura de tela de 2024-09-13 14-09-48

closes #1326

### DIFF
--- a/src/components/topics/TopicDisplay.tsx
+++ b/src/components/topics/TopicDisplay.tsx
@@ -22,14 +22,15 @@ const TopicDisplay = ({
     const [tags, setTags] = useState<any[]>([]);
     const { t } = useTranslation();
     const dispatch = useDispatch();
+    let timeout: NodeJS.Timeout;
 
     useEffect(() => {
         const inputValueFormatted = inputValue.map((inputValue) =>
             inputValue?.value
                 ? {
-                      label: inputValue?.label.toLowerCase().replace(" ", "-"),
-                      value: inputValue?.value,
-                  }
+                    label: inputValue?.label.toLowerCase().replace(" ", "-"),
+                    value: inputValue?.value,
+                }
                 : inputValue
         );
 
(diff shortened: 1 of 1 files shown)


thesocialdev:
Have you considered using MUI debounced input? https://mui.com/joy-ui/react-input/#debounced-input

LuizFNJ:
> Have you considered using MUI debounced input? https://mui.com/joy-ui/react-input/#debounced-input

In Aletheia, the Select component from Ant Design is used for selecting topics with native functionalities such as filtering and multiple selections. The Input from MUI is a simple text entry field with basic behavior.

The Select is used with styled-components, leveraging the power of styled-components to modify the appearance of an already existing and styled Ant Design component. In MUI, you can also apply style customizations, but this is typically done directly with MUI's styling API, without the need for an additional tool like styled-components.

**Pros of Substitution:**
MUI may be lighter and offer more flexibility in customization and behavior.

**Cons of Substitution:**
It would be necessary to manually replicate native selection and filtering functionalities, which could add complexity to the code.

**Conclusion:** 
I believe that implementing debounced functionality in the input using MUI may not add significant complexity to the code, but replicating the features of multiple selection and filtering using MUI could increase complexity.

thesocialdev (approved):

PR #1383, opened 2024-10-01
Correction on the component metrics and addition

Toxin4:
Correction on the component metrics and addition and addition of the message when there is no metric analysis
image
image


pepermao (approved):
LGTM: We could do a revamp on the empty content design, but let's leave that for later

PR #1384, opened 2024-10-02
Add readme in repository localConfig

LuizFNJ:
Guia em Português
[guia de configuração PT.pdf](https://github.com/user-attachments/files/17234255/guia.de.configuracao.PT.pdf)
Guia em Inglês
[Configuration guide EN.pdf](https://github.com/user-attachments/files/17234328/Configuration.guide.EN.pdf)


config/README-EN.md
```diff
@@ -0,0 +1,116 @@
+<h1 align="center">Platform Configuration Guide</h1>
+
+## Configuration Environment `./config`
+The configuration environment consists of a `public` directory, which will overwrite the original during the build process, and a TypeScript file for specific customizations and color definitions.
+
+In the build process, a script copies the root `public` directory to `dist/public`. After that, it checks if the `SVG`, `JSON`, `PNG` files and favicons in `config/public` are in their respective places, according to the original pattern and correctly named. If everything is correct, the files are overwritten in the `dist` folder.
+
+## SVG logo
+Placing the `SVG` logo in `config/public/images` with the standard name `logo1_white.svg` will overwrite the original during the build process.
```
pepermao, line 9: Question: Should we find a better standard file name ?

config/README-EN.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Favicon 
+The favicon consists of the following files: `android-chrome-192x192.png`, `android-chrome-512x512.png`, `apple-touch-icon.png`, `favicon-16x16.png`, `favicon-32x32.png`, `favicon.ico`, `site.webmanifest`. Just place them in `config/public` with the respective names, and they will be overwritten during the build process.
+
+## Locales
+In locales, there will be a PT folder for Portuguese and an EN folder for English. In this folder, all the `JSON` files to be overwritten will be stored.
+#### Exemplo:
```
pepermao, line 16: chore (non-blocking): Use English word `example`

config/README-EN.md
```diff
@@ -0,0 +1,116 @@
...
+{
+    "copyright": "AlumiaLab.org ©{{date}}",
+    "contactEmail": "contato@alumialab.org",
+    "platformInfoTittle": "Movimento Alumia Lab",
+    "adressStreet": "Av teste teste, Nº 000",
+    "adressZipcode":"0.000-000 - bairro - cidade/estado",
+    "legalRegistration":"00.000.cnpj/0000-00" 
+}
+```
+- `diretórioRaiz/public/pt/footer.json` - Original file:
```
pepermao, line 28: chore (non-blocking): replace "diretórioRaiz" to "rootDirectory"

config/README-EN.md
```diff
@@ -0,0 +1,116 @@
...
+`localConfig.schema.ts` and `tsconfig.json` are configuration files for the environment.
+
+### Colors
+If defined in `localConfig`, they will be rendered on the site; otherwise, the color defined in the `colors` component will be returned. It is not mandatory to define all colors; you can choose only those you prefer, and they should be in hexadecimal format `#FFFFFF`.
+- Currently, all these colors can be customized:
+```
+{
+theme: {
+        colors:{
+            bluePrimary: "",
```
pepermao, line 66: Question: @thesocialdev I wonder if we could use more generic names, the color prefix could confuse users if their color pallete is based on orange for example

config/README-EN.md
```diff
@@ -0,0 +1,116 @@
...
+}
+```
+
+### Footer - Header - Home
+Here, we customize specific points of the site, such as an array of links that return clickable social media icons; links for the Donation, Repository, and Statute buttons; and the choice to display or not display components by setting `true` to show and `false` to hide.
+```
+const localConfig: LocalConfig = {
+    theme: {},
+    footer: {
+        socialMedias: [ "https://www.instagram.com","url","url","url"], //cada "string" é um icone de rede social
```
pepermao, line 99: chore (blocking): use english tense

config/README-EN.md
```diff
@@ -0,0 +1,116 @@
...
+        showStatuteButton: {
+            show: true,
+            redirectUrl: "https://acessarestatuto.com",  //Altera o link do botão de estatuto
+        },
```
pepermao, line 103: chore (blocking): same to every other comment in portuguese

config/README-EN.md
LuizFNJ: I believe so, I would suggest "defaultLogo" what do you think?

pepermao: Sounds good, I'd recommend to use the same text case, e.g: `default_logo.svg`.

config/README-EN.md
thesocialdev: That's a good follow-up item we could work on. Let's create an issue.

pepermao (approved):

PR #1387, opened 2024-10-03
issues1170-addColorConditional

lucaslobatob:
Adicionar cor condicional para refletir a cor com base no ambiente do namespace no  UserEditRoles.
Screenshot from 2024-10-03 12-25-59
Screenshot from 2024-10-03 12-26-37

### DIFF
--- a/src/components/adminArea/Drawer/UserEditRoles.tsx
+++ b/src/components/adminArea/Drawer/UserEditRoles.tsx
@@ -24,6 +24,16 @@ const UserEditRoles = ({ currentUser, role, setUserRole, shouldEdit }) => {
             ? filterObjectByKeys(currentUser?.role, [nameSpace])
             : currentUser?.role;
 
+    const getBorderColor = (key) =>
+        nameSpace === NameSpaceEnum.Main
+    ? colors.bluePrimary 
+    : colors.blueSecondary;
+ 
+    const getRadioColor = (key) => 
+        nameSpace === NameSpaceEnum.Main 
+    ? colors.bluePrimary 
+    : colors.blueSecondary;
+
     const handleChangeRole = (value: Roles, key) => {
         setUserRole((role) => {
             return {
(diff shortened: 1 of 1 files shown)


src/components/adminArea/Drawer/UserEditRoles.tsx
```diff
@@ -52,7 +63,7 @@ const UserEditRoles = ({ currentUser, role, setUserRole, shouldEdit }) => {
                         <fieldset
                             style={{
                                 padding: 8,
-                                border: `1px solid ${colors.bluePrimary}`,
+                                border: `1px solid ${getBorderColor(role)}`,
```
pepermao, line 66: chore (block): Instead passing the role as parameter, use the `nameSpace` variable, which is the value we use to determine platform color

src/components/adminArea/Drawer/UserEditRoles.tsx
```diff
@@ -80,7 +91,11 @@ const UserEditRoles = ({ currentUser, role, setUserRole, shouldEdit }) => {
                                             disabled={!shouldEdit}
                                             key={role}
                                             value={role}
-                                            control={<Radio />}
+                                            control={<Radio 
+                                                style={{
```
pepermao, line 95: Do the same here

src/components/adminArea/Drawer/UserEditRoles.tsx
```diff
@@ -24,6 +24,16 @@ const UserEditRoles = ({ currentUser, role, setUserRole, shouldEdit }) => {
             ? filterObjectByKeys(currentUser?.role, [nameSpace])
             : currentUser?.role;
 
+    const getBorderColor = (key) =>
```
pepermao, line 27: Suggestion (non-blocking): Both `getBorderColor` and `getRadioColor` functions have the same purpose and return the same value, consider creating a unique function named `getComponentColor` to return their colors

pepermao (approved):

PR #1458, opened 2024-11-15
Disabling donation button in sidebar with LocalConfig

### LINKED ISSUE #1456: When disabling the donation button we should also remove it from the sidebar

LuizFNJ:
# Description
*I added the sidebar donation button to the LocalConfig conditional along with the other two home buttons*
Fixes #1456 
before:
Screenshot from 2024-11-15 12-06-33
after:
Screenshot from 2024-11-15 12-07-19

# Question
Screenshot from 2024-11-15 12-28-01
*When we disable the donation button on the home page and the user is not logged in, the sign up button remains in the same place when perhaps it should occupy the corner where the donation button was, does it make sense to resolve this in this PR?*


pepermao (changes requested):
~nitpick: Update the localConfig example and/or README~

pepermao (approved):

PR #1795, opened 2025-01-23
Fixing duplicate sources bug when creating claim

LuizFNJ:
# Description
**
Screenshot from 2025-01-23 03-53-58

Screenshot from 2025-01-23 03-53-06

Screenshot from 2025-01-23 03-53-27

Screenshot from 2025-01-23 03-53-41

Fixes #

# Type of change
- [x] Bug fix (non-breaking change which fixes an issue)     

# Testing
**

# Developer Checklist
## General
- [ ] Code is appropriately commented, particularly in hard-to-understand areas     
- [ ] Repository documentation has been updated (Readme.md) with additional steps required for a local environment setup.
- [ ] No `console.log` or related logging is added.
- [ ] No code is repeated/duplicated in violation of DRY. The exception to this is for new (MVP/Prototype) functionality where the abstraction layer may not be clear (comments should be added to explain the violation of DRY in these scenarios).   
- [ ] Documented with TSDoc all library and controller new functions

## Frontend Changes
- [ ] No new styling is added through CSS files (Unless it's a bugfix/hotfix)
- [ ] All types are added correctly

## Backend Changes
- [ ] All endpoints are appropriately secured with Middleware authentication
- [ ] All new endpoints have a interface schema defined

### Tests

- [ ] Enhancements or opportunities to improve performance, stability, security or code readability have been noted and documented in JIRA issues if not being addressed.     


server/claim/claim-revision/claim-revision.service.ts
```diff
@@ -204,17 +204,17 @@ export class ClaimRevisionService {
         if (sources && Array.isArray(sources)) {
             for (let source of sources) {
                 try {
-                    if (typeof source === "string") {
+                    const existingSource = await this.sourceService.findHref(source);
+                    if (existingSource && existingSource.length > 0) {
+                        for (let source of existingSource) {
```
caneppelevitor, line 209: nitpick: Consider renaming the variable inside the inner loop to avoid shadowing the outer source variable.

caneppelevitor, line 209: question: Would it be possible to optimize the loop execution by using Promise.all? Processing the updates concurrently instead of sequentially could potentially improve performance

caneppelevitor (changes requested):

server/source/source.service.ts
```diff
@@ -63,6 +63,10 @@ export class SourceService {
         return this.SourceModel.find({ match }, { _id: 1, href: 1 });
     }
 
+    findHref(href: string) {
```
caneppelevitor, line 66: nitpick: The current function name findHref suggests that it returns multiple results, but since we expect a single source, consider renaming it to getSourceByHref for better clarity. Additionally, updating the query to return a single object using .findOne() instead of .find()

server/claim/claim-revision/claim-revision.service.ts
caneppelevitor: question: Since we can only have one source with a given href, wouldn't it be better to simplify the logic and remove the inner loop? Instead of iterating over existingSource, we can directly check for its existence and update it.

caneppelevitor (changes requested):

caneppelevitor (approved):

caneppelevitor (approved):

thesocialdev (approved):

server/claim/claim-revision/claim-revision.service.ts
```diff
@@ -204,17 +204,15 @@ export class ClaimRevisionService {
         if (sources && Array.isArray(sources)) {
             for (let source of sources) {
                 try {
-                    if (typeof source === "string") {
+                    const existingSources = await this.sourceService.getSourceByHref(source);
+                    if (existingSources) {
+                            this.sourceService.updateTargetId(existingSources._id, claimId)
```
pepermao, line 209: nitpick: consider adding the await keyword in future opportunities

caneppelevitor: Awaiting updateTargetId is unnecessary here because its result isn't used, and it's the last operation in this branch, so it doesn't block further processing.

pepermao: The try/catch block won't catch possible errors when updating the source target ID if we don't await the response

caneppelevitor: Errors will still be caught by the try/catch block, even without await, since promise rejections propagate to the catch.

pepermao: Well, not sure about that. could you give me the documentation for that

pepermao: This is a very simple example where the console "save in the database" occurs even though we got an error in the db transaction. The promise execution starts but does not throw an error inside the try block because it's asynchronous. Any errors will be unhandled unless explicitly caught later.

```js
async function test() {
    setTimeout(() => {
        console.log('test')
        throw new Error("error")
    }, 2000)
}

try {
    test()
    console.log("save in the database")
} catch (e) {
    console.log("ERROR: ", e)
}
```

pepermao: https://javascript.info/async-await

caneppelevitor: After looking into this more deeply, I agree that await should be used for proper error handling. Using await ensures that if the underlying function—like updateTargetId—throws or returns a rejected Promise, the error will be caught in the same try/catch block. However, this only works if the function itself actually propagates errors by awaiting its own database calls. If, for instance, we never await source.save() inside updateTargetId, then a DB error in the save step won’t bubble up, even if we do await updateTargetId in the calling code. So, to fully capture errors, both the function and the caller need to consistently use await.

pepermao: Nice, could you provide the documentation that you looked at ? I think it's worth to share with everybody else. Thanks for the discussion, it's always good to have these kind of talks, I almost forgot about the try/catch error handling concept

PR #1799, opened 2025-01-25
Improving url validation regex

### LINKED ISSUE #1791: Enhance URL validation regex to support additional TLDs
The current URL validation regex is limited to a predefined set of TLDs, which may restrict sources from various regions. To ensure better flexibility and future-proofing, the regex should be improved to support a wider range of valid TLDs dynamically.

LuizFNJ:
# Description
*I limited the TLD to two characters from 'a' to 'z', so it accepts more options without restricting it to a set of TLDs. Additionally, I created a unit test to test some various URLs. The test uses a forEach on an array of predefined links, applying the URL pattern.*

*In conversation with @caneppelevitor, he conducted an in-depth investigation, and it was agreed that a change is needed in how we run Jest tests on the platform, as well as the possibility of separating the tests. Since this change may require a greater effort, it might make more sense to implement it in another PR.*

*In this PR, I kept the improvement of the pattern and encapsulated it in another component so that we can test only the URL validation functionality. I placed the test file in the appropriate location, alongside the component that contains this functionality, as it is independent and should not undergo changes after we fix and separate the unit tests on the platform.*

Fixes #1791 

# Type of change     
- [x] Existing feature enhancement (non-breaking change which modifies existing functionality)     

# Testing
*Create a font in the report and test with various types of TLDs.*

# Developer Checklist
## General
- [x] Code is appropriately commented, particularly in hard-to-understand areas     
- [x] Repository documentation has been updated (Readme.md) with additional steps required for a local environment setup.
- [x] No `console.log` or related logging is added.
- [x] No code is repeated/duplicated in violation of DRY. The exception to this is for new (MVP/Prototype) functionality where the abstraction layer may not be clear (comments should be added to explain the violation of DRY in these scenarios).   
- [x] Documented with TSDoc all library and controller new functions

### Tests
- [x] All existing unit and end to end tests pass across all services     
- [x] Unit and end to end tests have been added to ensure backend APIs behave as expected     

- [ ] Enhancements or opportunities to improve performance, stability, security or code readability have been noted and documented in JIRA issues if not being addressed.     


server/test-unit/urlValidation.spec.ts
```diff
@@ -0,0 +1,29 @@
+describe("URL Validation Tests", () => {
```
thesocialdev, line 1: praise: this is great! however, the test should be at the same level of the component file, with name `useFloatingLinkState.spec.ts`

that means we also need to make sure the test suite is also running for the `src` folder

thesocialdev (changes requested):

server/test-unit/urlValidation.spec.ts
LuizFNJ: *Currently, the tests are only run in the server folder. I can change this by simply modifying rootDir: "./" to rootDir: "../", or by creating another Jest configuration for the src folder. Which approach makes more sense?*

server/test-unit/urlValidation.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    const URL_PATTERN =
```
caneppelevitor, line 2: suggestion: Import the existing validateFloatingLink function instead of redefining the regex to ensure consistency and maintainability across the codebase.

server/test-unit/urlValidation.spec.ts
thesocialdev: we can go with a different path in the rootDir and if we need we change in the future

server/test-unit/urlValidation.spec.ts
LuizFNJ: I tried doing this, but it brought side effects from the other component. I tried using .only on the describe and it, but it still logged outputs from the other component, causing the test to fail. Is there another way to handle this, or should I just keep it as it is?

LuizFNJ: If I move the URL_PATTERN constant into the test and import it into the other places that are already using it, does that make sense? Or is that what you meant? @caneppelevitor

LuizFNJ:
## Question
*For some reason, I haven't investigated it thoroughly. When I run any test on the platform, even specifying to test only a specific component, it keeps showing "Running coverage on untested files..." and failing the test overall, but the specific test component I created continues to pass. I don't know why it runs untapped files, but should I create an issue for this?*
image

server/test-unit/urlValidation.spec.ts
LuizFNJ: *I think that by doing this it caused an error in the test-lint because apparently it started running more tests and returning more errors. makes sense?* @thesocialdev

caneppelevitor:
The issue on test-lint occurs because of the way Jest resolves modules when rootDir is changed in the Jest config.

When using:

`import * as mongoose from "mongoose";`

Jest treats mongoose as an ES module, which can cause mongoose.Types to become undefined in certain configurations.

The correct approach is to use:


`import mongoose from "mongoose";`

This ensures that the CommonJS default export is correctly resolved, avoiding issues when Jest's rootDir is modified.

Fix: Replace all instances of import * as mongoose from "mongoose"; with import mongoose from "mongoose"; throughout the project to ensure proper module resolution

thesocialdev (approved):

PR #1854, opened 2025-03-15
Migrating affix to MUI

### LINKED ISSUE #1558: Migrate component AffixPreviewButton.tsx - Import { Affix }

LuizFNJ:
# Description
*In this PR, I migrated the affix used in AffixPreviewButton.tsx and realized it was using the same logic I had already replicated for another affix. Therefore, I encapsulated the entire functionality into a new component and called it in both places.*

# Question 
@thesocialdev 
*I noticed that when the affix is used in the report drawer, the functionality of keeping the affix with a fixed position when the screen is scrollable is not working. Should I create an issue for this?*
*The affix used in the report is behind a feature flag and is currently turned off because it wasn’t working. I will create an issue to resolve this as well.*
*Should I create two separate issues or combine them into one?*

Fixes #1558
 
In the speech:
Screenshot from 2025-03-15 12-39-16
Screenshot from 2025-03-15 12-39-11
In the report:
Screenshot from 2025-03-15 12-38-51
Screenshot from 2025-03-15 12-38-44

# Testing
*To test the first affix, go to the claim page and test the component for showing and hiding the speech checks. The second use of the affix would be on the report page for the preview button of the report, but this functionality is currently disabled via a feature flag. In this PR, I only migrated the component and did part of the visual work.*

### DIFF
--- a/src/components/Collaborative/Components/AffixAletheiaButton.tsx
+++ b/src/components/Collaborative/Components/AffixAletheiaButton.tsx
@@ -1,0 +1,54 @@
+import { Box } from "@mui/material";
+import React, { useEffect, useRef, useState } from "react";
+
+interface AffixAletheiaButtonProps {
+  Children: React.ReactNode;
+  style?: React.CSSProperties;
+}
+
+const AffixAletheiaButton: React.FC<AffixAletheiaButtonProps> = ({ Children, ...style }) => {
+  const [isAffixed, setIsAffixed] = useState(false);
+  const ref = useRef<HTMLDivElement | null>(null);
+
+  useEffect(() => {
+    const handleScroll = () => {
+      if (!ref.current) return;
+
+      const { bottom } = ref.current.getBoundingClientRect();
+      const isFixed = bottom > window.innerHeight;
+
+      if (isFixed) {
+        setIsAffixed(true);
+      } else {
+        setIsAffixed(false);
+      }
+    };
+
+    if (ref.current) {
+      const scrollableSpeech = ref.current.scrollHeight > window.innerHeight;
+      if (scrollableSpeech) {
+        setIsAffixed(true);
+      }
+    }
+
+    window.addEventListener("scroll", handleScroll);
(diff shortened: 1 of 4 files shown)


LuizFNJ:
I couldn't think of a better name for the new component, so I kept it as AffixAletheiaButton.tsx. If you have a better suggestion. @thesocialdev

thesocialdev:
@LuizFNJ if it doesn't work in the drawer you should fix in this PR.

src/components/Collaborative/Components/AffixAletheiaButton.tsx
```diff
@@ -0,0 +1,54 @@
+import { Box } from "@mui/material";
+import React, { useEffect, useRef, useState } from "react";
+
+interface AffixAletheiaButtonProps {
+  Children: React.ReactNode;
+  style?: React.CSSProperties;
+}
+
+const AffixAletheiaButton: React.FC<AffixAletheiaButtonProps> = ({ Children, ...style }) => {
```
thesocialdev, line 9: suggestion: rename to `AffixButton`

thesocialdev (changes requested):

src/components/Collaborative/Components/AffixAletheiaButton.tsx
LuizFNJ: There is already a component with that name:

Screenshot from 2025-03-17 20-55-08

LuizFNJ:
> @LuizFNJ if it doesn't work in the drawer you should fix in this PR.

Fixed in drawer:

[Screencast from 03-17-2025 10:54:14 PM.webm](https://github.com/user-attachments/assets/2d999ab5-8892-4718-8bee-3ee31eececf4)

src/components/Collaborative/Components/AffixAletheiaButton.tsx
LuizFNJ: I left it in AffixButtonV2.tsx

thesocialdev (approved):

PR #1926, opened 2025-06-18
New feature to add namespaces in the user edit form

### LINKED ISSUE #1465: Enhancement: New feature to add user in NameSpaces via the UserEditForm
### Background Information

*During Aletheia's workshops with universities, it was noted that adding new users to specific namespaces proved challenging, as it required switching back and forth between the namespaces page and the admin page. This process hindered the efficiency of the admin.*

### Steps to Reproduce

*On the user edit page, there should be a feature to add namespaces to the user.*
*An example of where this new feature should fit:*
Untitled

### How

*Suggestion:*

*A functionality and design similar to adding topics but for adding NameSpaces to the user in UserEditorForm. 
It should follow the pattern of appearing only when the user clicks the +.*

*Example:*
Screenshot from 2024-11-20 19-22-32
Screenshot from 2024-11-20 19-23-36

### Expected Behavior

*A feature that allows the admin to add the user to specific namespaces within the UserEditForm page.*

LuizFNJ:
# Description
*In this PR, I added a new feature that allows assigning users to namespaces from the user edit page. Previously, this could only be done from the namespace page.*

*At first, I made changes to update the **keys** of the user's roles instead of just the values. This allowed us to add new namespaces to the current user while still being able to update the values of these keys (e.g., roles like `admin`, `regular`, etc.). This part worked quickly.*

*However, while testing, I realized it was also necessary to update the user ID inside each namespace where the user had been added. Previously, this wasn't done because the `updateUser` function only updated the user's roles — it didn’t add the user to the respective namespaces.*

*To fix this, I updated the `updateUser` function to also modify the relevant namespaces — adding or removing the user from them accordingly.*

Fixes #1465
Process adding user to namespace, changing the role within that namespace and entering the new namespace:
https://github.com/user-attachments/assets/eb98fafa-e708-4ad8-9bb0-4331d73f3550

# Testing
*To test this, try various update cases on the user edit page.*
*To access it, you need admin permissions. Go to the admin page, choose a user to edit, and test different update scenarios, including assigning namespaces to the user.*

### DIFF
--- a/server/auth/name-space/name-space.controller.ts
+++ b/server/auth/name-space/name-space.controller.ts
@@ -86,6 +86,12 @@ export class NameSpaceController {
     }
 
     @ApiTags("name-space")
+    @Get("api/name-space/")
+    async findAll() {
+        return await this.nameSpaceService.findAll();
+    }
+
+    @ApiTags("name-space")
     @Get("admin/name-spaces")
     public async adminNameSpaces(@Req() req: Request, @Res() res: Response) {
         const nameSpaces = await this.nameSpaceService.listAll();
--- a/server/auth/name-space/name-space.controller.ts
+++ b/server/auth/name-space/name-space.controller.ts
@@ -5,6 +5,7 @@ import {
     Param,
     Post,
     Put,
+    Query,
     Req,
     Res,
     UseGuards,
@@ -89,8 +90,14 @@ export class NameSpaceController {
     @UseGuards(AbilitiesGuard)
     @CheckAbilities(new AdminUserAbility())
     @Get("api/name-space/")
-    async findAll() {
-        return await this.nameSpaceService.listAll();
+    async findAllOrFiltered(
+        @Query("userId") userId?: object,
+    ) {
+        if (userId) {
+            return this.nameSpaceService.findByUser(userId);
+        }
+
+        return this.nameSpaceService.listAll();
     }
 
     @ApiTags("name-space")
--- a/server/auth/name-space/name-space.controller.ts
+++ b/server/auth/name-space/name-space.controller.ts
@@ -89,7 +89,7 @@ export class NameSpaceController {
     @ApiTags("name-space")
     @UseGuards(AbilitiesGuard)
     @CheckAbilities(new AdminUserAbility())
-    @Get("api/name-space/")
+    @Get("api/name-space")
     async findAllOrFiltered(
         @Query("userId") userId?: string,
     ) {
--- a/server/auth/name-space/name-space.controller.ts
+++ b/server/auth/name-space/name-space.controller.ts
@@ -94,7 +94,7 @@ export class NameSpaceController {
         @Query("userId") userId?: string,
     ) {
         if (userId) {
-            return this.nameSpaceService.findByUser(new Types.ObjectId(userId));
+            return this.nameSpaceService.findByUser(userId);
         }
 
         return this.nameSpaceService.listAll();
--- a/server/auth/name-space/name-space.controller.ts
+++ b/server/auth/name-space/name-space.controller.ts
@@ -86,9 +86,11 @@ export class NameSpaceController {
     }
 
     @ApiTags("name-space")
+    @UseGuards(AbilitiesGuard)
+    @CheckAbilities(new AdminUserAbility())
     @Get("api/name-space/")
     async findAll() {
-        return await this.nameSpaceService.findAll();
+        return await this.nameSpaceService.listAll();
     }
 
     @ApiTags("name-space")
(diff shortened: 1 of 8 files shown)


server/auth/name-space/name-space.controller.ts
```diff
@@ -85,6 +85,12 @@ export class NameSpaceController {
         return await this.nameSpaceService.update(id, newNameSpace);
     }
 
+    @ApiTags("name-space")
+    @Get("api/name-space/")
```
caneppelevitor, line 89: Security: This endpoint should be protected using our ability guard.

src/components/adminArea/Drawer/UserEditForm.tsx
```diff
@@ -21,11 +23,38 @@ const UserEditForm = ({ currentUser, setIsLoading }) => {
...
+                const selected = namespaces.filter((ns) =>
+                    userNamespaceKeys.includes(ns.slug)
+                );
+                setSelectedNamespaces(selected);
+
+            } catch (err) {
+                console.error("Erro ao buscar namespaces:", err);
+            }
+        };
+        if (currentUser?.role) {
```
caneppelevitor, line 52: Question: What exactly are we verifying here, is it whether the user has a specific role? Just noting that only admins should have access to the user edit page

src/components/adminArea/Drawer/UserEditForm.tsx
```diff
@@ -45,14 +74,29 @@ const UserEditForm = ({ currentUser, setIsLoading }) => {
         try {
             setIsLoading(true);
             const sendBadges = badges.map((badge) => badge._id);
+            const selectedSlugs = selectedNamespaces.map(ns => ns.slug);
+            const updatedRole = { ...role };
+
+            Object.keys(updatedRole).forEach((key) => {
```
caneppelevitor, line 80: Nitpick: Consider using "role" instead of "key" for better clarity and semantic meaning

caneppelevitor (changes requested):

src/components/adminArea/Drawer/UserEditForm.tsx
LuizFNJ: This was a check to see if there are roles, i.e. {"namespace": "role"}. rethinking the flow better and this check really doesn't make sense because after all the namespaces must always exist and otherwise it would be a user data error.

server/users/users.service.ts
```diff
@@ -144,11 +149,36 @@ export class UsersService {
...
+                if (!currentId.includes(ns._id)) {
+                    await this.NameSpaceModel.updateOne(
+                        { _id: ns._id },
+                        { $addToSet: { users: user._id } }
+                    );
+                }
+            }
+    
+            for (const ns of currentNamespace) {
+                if (!namespaceSlugs.includes(ns.slug) && !newIds.includes(ns._id)) {
```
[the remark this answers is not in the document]
LuizFNJ: It's not necessary, when I placed the correct check I forgot to remove the wrong one

src/components/adminArea/Drawer/UserEditForm.tsx
```diff
@@ -45,14 +70,27 @@ const UserEditForm = ({ currentUser, setIsLoading }) => {
...
+                if (key !== "main" && !selectedSlugs.includes(key)) {
```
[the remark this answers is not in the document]
caneppelevitor: issue: Still not using the nameSpace enum.

src/api/namespace.ts
```diff
@@ -32,12 +32,16 @@ const updateNameSpace = (nameSpace, t) => {
         });
 };
 
-const getNameSpaces = () => {
+const getNameSpaces = (param?: string) => {
+    const params: Record<string, string> = {};
+
+    if (typeof param === "string" && param !== null) {
```
caneppelevitor, line 38: nitpick: Could we make this verification outside the helper? Maintaining the business logic on the component .

src/components/adminArea/Drawer/UserEditForm.tsx
```diff
@@ -21,11 +23,32 @@ const UserEditForm = ({ currentUser, setIsLoading }) => {
...
         setBadges(newValue);
     };
 
+    const handleChangeNameSpaces = (_event, newValue: NameSpace[]) => {
+        setSelectedNamespaces(newValue);
+    };
+    useEffect(() => {
+        const fetchNamespaces = async () => {
+            try {
+                const namespaces = await NameSpacesApi.getNameSpaces();
```
caneppelevitor, line 39: issue: Right now, if NameSpacesApi.getNameSpaces() fails, the catch {} will silently swallow the error and you’ll have no idea why your UI might be stuck empty

caneppelevitor (changes requested):

caneppelevitor:
Just Fix the last code smells and then it can be merged

PR #1930, opened 2025-07-03
Logout user if not found in MongoDB after Ory login

### LINKED ISSUE #1445: When the user is able to login but doesn't exist in the DB the session should be ended and the user logged out
### Background Information

_No response_

### Steps to Reproduce

_No response_

### How

_No response_

### Expected Behavior

_No response_

lucaslobatob:
# Description
*This PR implements an additional check in `SessionGuard` that ensures that after a successful login via Ory, the user also exists in the database (MongoDB).
If the `user_id` coming from Ory does not match a valid `_id` in the user collection, the session is terminated and the user is redirected to the `/signup-invite` route and creation.*

Fixes #1445 

# Type of change
- [X] Bug fix (non-breaking change which fixes an issue)     

# Testing
*Steps:
1. Delete a user that is registered in your database and the ory;
2. Enter /login;
3. Fill in the inputs with the data of this user that is not in your database.
4. By clicking on "Entrar", you will be informed that the login was successful, but you will be redirected to the /signup-invite route, inviting you to register in the environment you are in.*

# Developer Checklist
## General

## Frontend Changes
- [X] No new styling is added through CSS files (Unless it's a bugfix/hotfix)
- [X] All types are added correctly

## Backend Changes
- [X] All endpoints are appropriately secured with Middleware authentication
- [X] All new endpoints have a interface schema defined

### Tests

### Test IDs
     


server/app.module.ts
```diff
@@ -173,6 +174,16 @@ export class AppModule implements NestModule {
                     provide: APP_GUARD,
                     useExisting: NameSpaceGuard,
                 },
+                {
+                    provide: APP_GUARD,
+                    useClass: SessionOrM2MGuard,
```
thesocialdev, line 179: question: why is this needed? The guards are already declared above.

server/auth/session.guard.ts
```diff
@@ -21,6 +47,13 @@ export class SessionGuard extends BaseGuard {
         const request = httpContext.getRequest();
         const response = httpContext.getResponse();
         const type = this.configService.get<string>("authentication_type");
+        const baseUrl = `http://${request.headers.host}`;
```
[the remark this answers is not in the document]
caneppelevitor: So I think the issue here is that our appAffiliation step, since it runs before verifying the user in our database, is overriding our redirect. That’s why this change was necessary. We need to verify the user in the DB before running appAffiliation; that way, we can use the checkAndRedirect function.

Let’s clarify this discussion in the daily.

server/auth/session.guard.ts
```diff
@@ -42,24 +74,24 @@ export class SessionGuard extends BaseGuard {
...
                         request,
                         response,
                         isPublic,
                         "/unauthorized"
                     );
                 }
+
+                const mongoUserId = session?.identity?.traits?.user_id;
+                try {
+                    await this.usersService.getById(mongoUserId);
```
thesocialdev, line 88: suggestion: a better architecture is to always return something from a function, including errors, and treat them where you call it.

server/auth/name-space/name-space.guard.ts
```diff
@@ -3,13 +3,14 @@ import {
...
     NotFoundException,
+    Scope,
     UnauthorizedException,
 } from "@nestjs/common";
 import { ConfigService } from "@nestjs/config";
 import { Configuration, FrontendApi } from "@ory/client";
 import { NameSpaceService } from "./name-space.service";
 
-@Injectable()
+@Injectable({ scope: Scope.REQUEST })
```
thesocialdev, line 13: suggestion: I think this is a good change, but make sure this is will tested across all endpoints.

server/auth/m2m-or-session.guard.ts
```diff
@@ -1,8 +1,13 @@
...
+    Injectable,
+    ExecutionContext,
+    CanActivate,
+    Scope,
+} from "@nestjs/common";
 import { SessionGuard } from "./session.guard";
 import { M2MGuard } from "./m2m.guard";
 
-@Injectable()
+@Injectable({ scope: Scope.REQUEST })
```
thesocialdev, line 10: same as below

thesocialdev (changes requested):

server/auth/session.guard.ts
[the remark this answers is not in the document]
caneppelevitor: I still have my concerns if this verification is still necessary now that we are checking the user on mongo before the app affiliation

PR #1941, opened 2025-07-21
fix: correct isDifferentLabel function in cross-checking workflow

thesocialdev:
## Summary
- Fix bug where fact-checkers lost access to the review form after clicking "Adicionar Comentário" during cross-checking
- Correct the `isDifferentLabel` function in the state machine workflow to use the correct context path
- Add comprehensive Cypress tests for cross-checking comment workflow

## Root Cause
The bug was in `src/machines/reviewTask/machineWorkflow.ts:17` where the `isDifferentLabel` function was incorrectly checking `context?.classification` instead of `context?.reviewData?.classification`. This caused the state machine transition conditions to fail, leaving the form in an inaccessible state.

## Why Admins Were Unaffected
Admin users have UI-level bypasses that allow them to see the form and buttons regardless of state machine state, plus they can reset workflows using the `reAssignUser` action.

## Changes Made
1. **Fixed State Machine Guard**: Updated `isDifferentLabel` function to use correct context path
2. **Added Cypress Tests**: Created comprehensive e2e tests that validate:
   - Form remains accessible after adding comments during cross-checking
   - State transitions work correctly for both matching and different label scenarios
   - Cross-checking workflow completes successfully

## Test Plan
- [x] Fix applied to `machineWorkflow.ts`
- [x] Cypress tests added in `cypress/e2e/tests/cross-checking.cy.ts`
- [x] Tests validate the bug fix works correctly
- [x] Tests ensure no regression in cross-checking workflow

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/cypress/tsconfig.json
+++ b/cypress/tsconfig.json
@@ -4,5 +4,5 @@
       "lib": ["es5", "dom"],
       "types": ["cypress"]
     },
-    "include": ["**/*.ts", "../src/components/Home/CTAButton.cy.ts"]
+    "include": ["**/*.ts"]
   }


caneppelevitor:
Question: Are we removing the Cypress tests from the pipeline? If that’s the case, what’s the point of creating them for cross‑checking if we’re not using them anymore?

caneppelevitor (approved):

caneppelevitor (approved):

PR #2006, opened 2025-09-15
Prevent duplicate sources by reusing existing data_hash

### LINKED ISSUE #1967: Bug: Duplicated Sources on VerificatioRequest
When creating two different verificationRequests with the same source we get an error of duplicated datash on the data base

## Expected behaviour
If the source already exists, fetch its information and save it into the VerificationRequest record.

lucaslobatob:
# Description
*This PR adds a check to the SourceService to prevent duplicates. Now, when a user enters a link that already exists, the service generates the data_hash from the href and checks whether a Source is already registered. If it exists, the existing record is returned; otherwise, a new document is created. This ensures that there are no multiple Sources for the same link.*

Fixes #1967 

# Type of change
- [X] Bug fix (non-breaking change which fixes an issue)     

# Testing
*Steps:*
1. Log in platform.
2. Create a new Verification Request and add one source.
3. Create a new Verification Request and add the same source as the previous or existing Verification Request.

Note: Previously, the platform prevented you from placing an existing source already in our database; now, as we can see in the image, the Verification Request is created with the same source.

Screenshot from 2025-09-15 14-40-54

# Developer Checklist
## General

## Backend Changes
- [X] All endpoints are appropriately secured with Middleware authentication
- [X] All new endpoints have a interface schema defined

### Tests

### Test IDs
     

### DIFF
--- a/server/source/source.service.ts
+++ b/server/source/source.service.ts
@@ -59,7 +59,6 @@ export class SourceService {
             return existingSource;
         }
 
-        //TODO: don't create duplicate sources in one claim review task
         return await new this.SourceModel(data).save();
     }
 
--- a/server/source/source.service.ts
+++ b/server/source/source.service.ts
@@ -52,7 +52,7 @@ export class SourceService {
         data.user = Types.ObjectId(data.user);
 
         const existingSource = await this.SourceModel.findOne({
-            data_hash: data.data_hash,
+            data_hash: { $eq: data.data_hash },
         });
 
         if (existingSource) {
--- a/server/source/source.service.ts
+++ b/server/source/source.service.ts
@@ -38,6 +38,15 @@ export class SourceService {
         }
         data.data_hash = md5(data.href);
         data.user = Types.ObjectId(data.user);
+
+        const existingSource = await this.SourceModel.findOne({
+            data_hash: data.data_hash,
+        });
+
+        if (existingSource) {
+            return existingSource;
+        }
+
         //TODO: don't create duplicate sources in one claim review task
         return await new this.SourceModel(data).save();
     }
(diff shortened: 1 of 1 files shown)


server/source/source.service.ts
```diff
@@ -1,7 +1,12 @@
-import { Injectable, NotFoundException } from "@nestjs/common";
+import {
+    BadRequestException,
+    Injectable,
+    NotFoundException,
+} from "@nestjs/common";
 import { Model, Types } from "mongoose";
 import { SourceDocument, Source } from "./schemas/source.schema";
 import { InjectModel } from "@nestjs/mongoose";
+import validator from "validator";
```
caneppelevitor, line 9: question: this `validador` is part of our dependencies and could we use the already used on our code base `class-validator` instead?

caneppelevitor (changes requested):

server/source/source.service.ts
lucaslobatob: We could use class-validator for this, but in this case validator provides a stronger and more specific URL validation with options like require_protocol. Since the validation happens inside the service (and not only at DTO level), validator is a better fit here.

server/source/source.service.ts
```diff
@@ -36,8 +41,24 @@ export class SourceService {
...
+
+        const existingSource = await this.SourceModel.findOne({
+            data_hash: { $eq: data.data_hash },
+        });
+
+        if (existingSource) {
+            return existingSource;
+        }
+
         //TODO: don't create duplicate sources in one claim review task
```
caneppelevitor, line 62: Now we can remove this TODO comment

caneppelevitor (changes requested):

caneppelevitor (approved):

PR #2339, opened 2026-03-23
chore: Update Next from v12 to v15

lucaslobatob:
# Description
This PR updates Next.js from version 12.1.0 to 15.5.14, temporarily maintaining the Pages Router-based architecture. Migration to App Router will be addressed in a future PR, aiming to reduce risks and facilitate incremental validation of changes. The eslint-config-next file was also updated from v12 to v13, reflecting the changes recommended by the latest Next ecosystem.

During the update, some settings were added to tsconfig.json, such as `strictNullChecks`, `useUnknownInCatchVariables`, and `strictPropertyInitialization`, which were temporarily disabled. This decision was made to prevent existing typing inconsistencies in the project from blocking the build after the upgrade. The idea is to gradually review and correct these typing issues to reactivate strict mode in the future.

Another important point was the removal of the `.babelrc` file and the configuration of styled-components directly in `next.config.js` using the native Next compiler. This leverages the use of SWC, bringing performance gains and simplifying project configuration, eliminating the need for Babel.

Also, the `@IsDate` decorator was changed to `@IsDateString` in the verification request creation DTO, ensuring compatibility with the actual format of the data sent by the API, which are strings in the ISO standard, preventing validation failures.

The navigation flow was refactored, moving the API's router.push to the CreateVerificationRequestView component. This removes the API-bound navigation and allows it to be controlled directly by the UI layer.

Finally, in view.service, the import of createServer was changed from import to require. This change was necessary to ensure compatibility with the current runtime environment, avoiding potential issues related to the difference between ESModules and CommonJS modules after the Next update, especially in server-side contexts.

Related Ticket #2326 

# Type of change
- [X] Existing feature enhancement (non-breaking change which modifies existing functionality)     

# Developer Checklist
## General
- [X] Code is appropriately commented, particularly in hard-to-understand areas

### Tests

### Test IDs
     


src/components/VerificationRequest/verificationRequestForms/CreateVerificationRequestView.tsx
```diff
@@ -32,9 +33,14 @@ const CreateVerificationRequestView = () => {
         };
 
         verificationRequestApi
-            .createVerificationRequest(t, router, newVerificationRequest)
+            .createVerificationRequest(t, newVerificationRequest)
             .then((s) => {
```
LuizFNJ: nit: try to use a more descriptive nomenclature for 's'

src/components/BackButton.cy.tsx
```diff
@@ -1,62 +1,90 @@
...
+        off: cy.stub(),
+        emit: cy.stub(),
+    },
+    isReady: true,
+    basePath: "",
+    isLocaleDomain: false,
+    ...overrides,
+});
+
+const mountWithRouter = (component: React.ReactNode, routerOverrides = {}) => {
```
LuizFNJ: Suggestion: Move mountWithRouter to a global Cypress command. This will allow us to call cy.mountWithRouter() in future tests, centralizing the Next.js Router mock logic and following DRY principles.

server/tsconfig.json
```diff
@@ -11,8 +11,13 @@
         "sourceMap": true,
         "outDir": "../dist/",
         "rootDir": "../",
-        "baseUrl": "./",
-        "incremental": true
+        "incremental": true,
```
LuizFNJ, line 14: I think if we lower the TypeScript strictness, it would be better if this PR doesn't go to staging yet, since we’ll eventually have to migrate the whole structure in a new PR—and we don't know how long that will take. It would be unfeasible to have low strictness for new tasks that come up during the process. I’d suggest that this PR be reviewed and approved but kept open as a base for the next one. By branching the new PR off this infra one, we could finalize everything there and then move to staging without lowering the TypeScript strictness. Does this make sense, or should we take another approach?

@caneppelevitor @lucaslobatob

server/verification-request/dto/create-verification-request-dto.ts
```diff
@@ -36,10 +42,10 @@ export class CreateVerificationRequestDTO {
     @IsOptional()
     email: string;
 
-    @IsDate()
+    @IsDateString()
     @ApiProperty()
     @IsOptional()
-    date: Date;
+    date: string;
```
caneppelevitor: issue: Changing `date` from `Date` to `string` fixes validation, but any downstream code calling Date methods on this field (`.getTime()`, `.toISOString()`, etc.) will break at runtime with no compile error since `strict` is off.

  Worth a quick grep:
  git grep -n '\.date\.' server/verification-request/

  If anything treats it as a Date object, add a `new Date(dto.date)` conversion at the service layer.

package.json
```diff
@@ -217,7 +217,7 @@
         "dotenv": "^16.4.5",
         "env-cmd": "^10.1.0",
         "eslint": "^9.39.2",
-        "eslint-config-next": "^12.0.10",
+        "eslint-config-next": "13",
```
caneppelevitor: question: are we able to bump this to follow the next version ? `eslint-config-next` is at v13 but Next is 15.5.14. This means ESLint won't flag Next 15-specific issues (async dynamic params, new import patterns, deprecated APIs). If possible bump to `"15"` to match.

caneppelevitor (changes requested):

PR #2367, opened 2026-04-03
fix: security hardening, TypeScript typing improvements, and error handling cleanup

caneppelevitor:
## Summary 

Related Ticket  #2450

### Security
- Remove `@Public()` decorator accidentally left on unattributed claim endpoint
- Sanitize all MongoDB `$regex` queries with shared `escapeRegex` utility to prevent NoSQL injection
- Add `@IsEmail()` validation to `CreateUserDTO`

### Error Handling
- Replace `return error` pattern in claim controllers with proper exception propagation
- Fix empty `.catch()` in source controller — convert to async/await
- Fix unawaited promises in `comment.service.ts` (`forEach` → `Promise.all`)
- Fix mixed async/promise patterns in `users.controller.ts`
- Add safe optional chaining on `req.user.role` in debate claim creation

### TypeScript
- Add response types and parameter types to 7 core API files (claim, claimReview, source, reviewTask, personality, user)
- Eliminate all 15 `@ts-ignore` comments with proper type solutions
- Add return types to claim and claim-review service methods
- Create shared `ApiResponse.ts` with `PaginatedResponse<T>`, `TranslationFn`, and other reusable types

### Validation
- Add `UpdateClaimReviewDTO` with class-validator decorators
- Add `UpdateHiddenStatusDTO` for claim visibility endpoint
- Add `CreateCommentDTO`, `UpdateCommentDTO`, `DeleteReplyDTO` for comment endpoints

### Bug Fixes
- Fix `SelectOptions` passing event object instead of input value to user search (`searchName=[object Object]`)
- Add `ErrorBoundary` component for feature-level use

## Test plan
- [ ] Verify claim creation flows (speech, image, debate, unattributed)
- [ ] Verify claim-review hide/unhide validates payloads correctly
- [ ] Verify comment endpoints validate input via new DTOs
- [ ] Verify user and personality search work with regex sanitization
- [ ] Verify user assignment dropdown returns correct search results
- [ ] Verify reCAPTCHA renders on review task forms
- [ ] Run unit, e2e, and Cypress test suites


server/claim-review/claim-review.controller.ts
```diff
@@ -69,7 +70,10 @@ export class ClaimReviewController {
     @AdminOnly()
     @ApiTags("claim-review")
     @Put("api/review/:id")
-    async update(@Param("id") reviewId, @Body() body): Promise<any> {
+    async update(
+        @Param("id") reviewId: string,
+        @Body() body: UpdateClaimReviewDTO
+    ): Promise<unknown> {
```
LuizFNJ: Question: Setting the return type to unknown here was intentional? While it decouples the Controller from the Service's internal types, we lose Type Safety on the API response. Is there a reason we aren't using UpdateWriteOpResult here?

server/claim/claim.controller.ts
```diff
@@ -82,8 +83,7 @@ export class ClaimController {
         }
         if (query.personality && !query.isHidden) {
             inputs.personalities = new mongoose.Types.ObjectId(
-                // @ts-ignore
-                query.personality
+                query.personality as string
```
LuizFNJ, line 86: Nitpick: It would be better to type the parameter as GetClaimsDTO instead of relying on type casting with as string inside the function

```typescript
_verifyInputsQuery(query: GetClaimsDTO) {....}
```

src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,84 @@
...
+                >
+                    <Typography
+                        variant="h5"
+                        style={{
+                            color: colors.primary,
+                            fontWeight: 600,
+                            marginBottom: "16px",
+                        }}
+                    >
+                        Something went wrong
```
LuizFNJ: Nitpick: These strings should be internationalized using next-i18next.

src/types/ApiResponse.ts
```diff
@@ -0,0 +1,21 @@
...
+    title: string;
+    path: string;
+}
+
+export interface PasswordChangeResponse {
+    success: boolean;
+    message: string;
+}
+
+export type TranslationFn = (
```
LuizFNJ: Nitpick: Why not use the official TFunction type from i18next instead of defining a custom TranslationFn?

server/auth/name-space/name-space.guard.ts
LuizFNJ, line 40: If you type the return of this findOne as Promise<NameSpaceDocument> and use .exec():

```typescript
findOne(match): Promise<NameSpaceDocument> {
    return this.NameSpaceModel.findOne(match).exec();
}
```

And inside the NameSpace schema, type the users as UserDocument[]. We need UserDocument specifically because User doesn't have the _id explicitly defined, and this will remove the need for that double 'as' casting... see the next comment.

server/auth/name-space/name-space.guard.ts
```diff
@@ -46,8 +46,10 @@ export class NameSpaceGuard implements CanActivate {
             }
 
             const userHasAccess = namespace.users.some(
```
LuizFNJ, line 48: After fixing the types to UserDocument[], you can get rid of the double casting. It becomes much cleaner:

```typescript
const userHasAccess = namespace.users.some(
    (user) => user._id.toString() === user_id
);
```

Does that make sense?

server/badge/badge.controller.ts
```diff
@@ -87,15 +87,20 @@ export class BadgeController {
                 this.usersService.updateUser(user._id, {
```
LuizFNJ, line 87: If we update the `updateUser` signature to include the `Promise<User>` return like this:

```typescript
async updateUser(
    userId,
    updates: { role?: object; badges?: Badge[]; state?: Status }
): Promise<User> { ... }
```

And then, in the User schema, change badges: Badge[] to badges: BadgeDocument[], the _id property will be natively recognized. This way, we fix the type safety at the source and avoid the double casting in the filters... see the next comment.

server/badge/badge.controller.ts
```diff
@@ -87,15 +87,20 @@ export class BadgeController {
...
                     badges: user.badges.filter(
                         (userBadge) =>
-                            // @ts-ignore
-                            userBadge._id.toString() !==
-                            updatedBadge._id.toString()
+                            (
+                                userBadge as unknown as {
```
LuizFNJ: After fixing the types to `BadgeDocument[]` as suggested in the previous comment, we can get rid of the double casting entirely. The logic becomes much cleaner and easier to read:

```typescript
if (!isUserIdOnUsersList) {
                this.usersService.updateUser(user._id, {
                    badges: user.badges.filter(
                        (userBadge) => userBadge._id.toString() !== updatedBadge._id.toString()
                    ),
                });

                user.badges = user.badges.filter(
                    (userBadge) => userBadge._id.toString() !== updatedBadge._id.toString()
                )
```

Does that make sense?

server/claim/claim.controller.ts
caneppelevitor: The as string cast here is a simple type narrowing on a private helper, introducing a DTO for an internal method would add unnecessary overhead since the input is already validated upstream.

src/components/ErrorBoundary.tsx
caneppelevitor: My bad I removed the usage of this component in a previous commit, deleted this file.

LuizFNJ (approved):

PR #2389, opened 2026-04-13
Fix session expiry handling and error page routing

caneppelevitor:
## Summary
- Fix 500 errors when session expires during review task workflows (save-draft, user search)
- Add session expired modal that redirects users back to their page after re-login
- Fix exception filters to properly route errors (API gets JSON, pages get redirected to 404/error pages)
- Add diagnostic context to error logs for better observability
- Migrate review task workflow API files to shared Axios factory with 401 handling
- Fix notification token requests with undefined user ID

## Follow-up
This PR fixes the immediate production errors and establishes the pattern for proper session expiry handling. The following items are planned as follow-up work to fully apply this pattern across the app:

- Migrate remaining ~20 API files to the shared Axios factory (`createApiInstance`)
- Add `req.user` null guards to other backend services that access user context without validation
- Create a dedicated `/500` error page (currently server errors redirect to `/404`)
- Normalize inconsistent error handling across API files (silent catches, missing feedback)

Related Ticket #2391

## Test plan
- [x] `curl` save-draft and user search without session return 401 (not 500)
- [x] Session expiry during active work shows modal, clicking "Log in" redirects to `/login?return_to=...`
- [x] After re-login, user returns to the same page
- [x] Non-API 404/500 errors redirect to the 404 page
- [x] No `/api/notification/token/undefined` requests in network tab
- [x] Server logs include userId, userAgent, and query params on errors

### DIFF
--- a/server/filters/http-exception.filter.ts
+++ b/server/filters/http-exception.filter.ts
@@ -115,9 +115,9 @@ export class AllExceptionsFilter implements ExceptionFilter {
                     message: safeMessage,
                 });
             } else if (status === HttpStatus.NOT_FOUND) {
-                response.status(status).redirect("/404");
+                response.redirect("/404");
             } else if (status === HttpStatus.INTERNAL_SERVER_ERROR) {
-                response.status(status).redirect("/404");
+                response.redirect("/404");
             } else {
                 response.status(status).json({
                     requestId,
--- a/server/filters/http-exception.filter.ts
+++ b/server/filters/http-exception.filter.ts
@@ -105,28 +105,22 @@ export class AllExceptionsFilter implements ExceptionFilter {
         // Send error response only if headers haven't been sent
         if (!response.headersSent) {
             const isApiRequest = request.originalUrl.startsWith("/api");
+            const shouldRedirect = [
+                HttpStatus.NOT_FOUND,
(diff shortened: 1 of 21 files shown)


server/filters/http-exception.filter.ts
```diff
@@ -79,13 +104,29 @@ export class AllExceptionsFilter implements ExceptionFilter {
 
         // Send error response only if headers haven't been sent
         if (!response.headersSent) {
```
LuizFNJ, line 106: Could it be cleaner by applying DRY? Like this:

```typescript
if (!response.headersSent) {
    const isApiRequest = request.originalUrl.startsWith("/api");

    const shouldRedirect = [HttpStatus.NOT_FOUND, HttpStatus.INTERNAL_SERVER_ERROR].includes(status);

    if (!isApiRequest && shouldRedirect) {
        return response.redirect("/404");
    }

    const errorResponse = {
        requestId,
        statusCode: status,
        timestamp: new Date().toISOString(),
        path: request.url,
        message: safeMessage,
    };

    response.status(status).json(errorResponse);
}
```

caneppelevitor: done

LuizFNJ (approved):

PR #2559, opened 2026-07-18
feat(mcp): remote MCP server with OAuth login and fact-checking tools

thesocialdev:
# Description

Adds a **remote MCP (Model Context Protocol) server** to the NestJS app, exposed at `POST /server/mcp`. Any MCP client (claude.ai connectors, Claude Code, Cursor) can connect, log in as a **real Aletheia user** via OAuth 2.1 in the browser (the "Notion connector" flow), and run fact-checking tools that call our existing domain services under the user's real role.

The entire feature is gated behind a new `mcp.enabled` config flag (**default `false`**), so this PR is inert until explicitly enabled per environment.

Design & plan committed under `docs/superpowers/specs/` and `docs/superpowers/plans/`.

Related Ticket # (issue) — _n/a (spike)_

## How it works

- **Endpoint:** `server/mcp/` module mounted at `/server/mcp` (path chosen to avoid the Next.js `:namespace?` view routes). Stateless Streamable HTTP transport — a fresh MCP server is built per request; tools call existing services directly (no internal HTTP hops).
- **Login (OAuth 2.1 / MCP auth spec):** unauthenticated requests get a `401` + `WWW-Authenticate` challenge pointing at RFC 9728 metadata (`/.well-known/oauth-protected-resource/server/mcp`), which names Ory as the authorization server. The client runs the browser authorization-code + PKCE flow against Ory, then calls tools with `Authorization: Bearer <token>`.
- **`McpAuthGuard`:** introspects the bearer token via Hydra. Machine tokens (`sub === client_id`) keep the existing `integration` behavior; **user** tokens are resolved via a new `OryService.getIdentity` to the real Kratos identity + role, so CASL/namespace rules apply as on the web. Tokens are cached (SHA-256-keyed, 60s TTL, lazy eviction); affiliation and identity-state checks fail closed.
- **Tools (17):** read — `search`, `list_claims`, `get_claim`, `list_personalities`, `get_personality`, `get_claim_review`, `list_review_tasks`, `get_review_task`, `list_verification_requests`, `get_verification_request`, `search_topics`, `list_sources`; write — `create_verification_request`, `add_review_comment`, `create_claim`, `create_personality`, `create_source`. Inputs are Zod-validated.

## ⚠️ Prerequisites before enabling (`mcp.enabled: true`)

Per environment, in the Ory Network console: enable **OAuth2 (Hydra)**, enable **Dynamic Client Registration**, and use Ory's **hosted login/consent**. Set `mcp.public_url` (and optionally `mcp.authorization_server`). See `server/mcp/README.md`.

# Type of change

- [x] New feature (non-breaking change which adds functionality)

# Testing

- Unit: `yarn test:unit` — `McpAuthGuard` spec (12 tests: machine/user tokens, affiliation fail-closed, inactive identity, token-cache hashing + TTL eviction) and `OryService.getIdentity` spec.
- E2E: `yarn env-cmd --silent vitest run --project e2e server/tests/mcp.e2e.spec.ts` — 11 tests: RFC 9728 metadata, MCP `initialize` handshake, GET→405, `tools/list` (all 17), `list_claims` payload, and the authorization suite (cross-namespace read reject, allowed-namespace pass, Regular-user write block, admin write allow, M2M write reject, `get_claim_review` param advertisement).
- `yarn build-ts` clean.
- Manual: MCP Inspector / `curl` against local dev (see README) — 401 challenge + metadata verified without auth.

**Note on CI:** `server/tests/claim-review.e2e.spec.ts` has 5 failures that are **pre-existing on `stage`** (verified by stashing this branch's changes — they fail identically without it), unrelated to this PR.

# Developer Checklist
## General
- [x] Code is appropriately commented, particularly in hard-to-understand areas
- [x] Repository documentation has been updated (`server/mcp/README.md` — setup, Ory prerequisites, connection guide)
- [x] No `console.log` added (uses Nest `Logger`)
- [x] No code repeated/duplicated in violation of DRY
- [x] Documented new library/controller functions

## Backend Changes
- [x] All endpoints are appropriately secured with Middleware authentication (`McpAuthGuard`; write tools additionally gated by namespace membership + write role)
- [x] All new endpoints have an interface schema defined (Zod input schemas per tool; RFC 9728 metadata shape)

### Tests
- [x] Unit and e2e tests added for backend APIs
- [ ] All existing unit and e2e tests pass across all services — _yes except 5 pre-existing `claim-review.e2e` failures on `stage`, unrelated to this PR_

# Notes for reviewers

- Built iteratively with per-task spec + adversarial code review. The security-sensitive pieces (`McpAuthGuard`, namespace/role authorization on tools) went through extra hardening rounds.
- **Known follow-ups tracked separately** (non-blocking, do not affect the default-off flag): (1) `get_review_task` / `add_review_comment` by-id namespace hardening; (2) a pre-existing `SessionGuard` affiliation fail-open (same class fixed here for the MCP guard); (3) admin/user-management tools and review-task state transitions were intentionally out of scope for v1.


server/auth/base.guard.ts
```diff
@@ -48,6 +48,8 @@ export abstract class BaseGuard implements CanActivate {
             "/api/user/register",
             "/api/claim", // Allow this route to be public temporarily for testing
             "/signup-invite",
+            "/server/mcp",
```
thesocialdev: If it needs authentication why an allowed public URL?

server/mcp/tools/schemas.ts
```diff
@@ -0,0 +1,132 @@
+import { z } from "zod";
```
thesocialdev, line 1: This file is no-no, the schemas should be derived of the entities or canonical zod objects, this will create drift between types.

server/auth/base.guard.ts
thesocialdev: Fixed in 48bf5b05. You're right — an authenticated route didn't belong in `isAllowedPublicUrl`. Root cause: the entry only existed to stop the session guard from 302-redirecting anonymous clients to `/login` (MCP needs a 401 OAuth challenge), and it was load-bearing because dev sets `override_public_routes=true`. The guard already lets every `/api/*` route bypass that redirect unconditionally, so I moved the endpoint to `/api/mcp` and removed the `/server/mcp` entry entirely. The only MCP-related entry left in the list is `/.well-known/oauth-protected-resource`, which is genuinely public (RFC 9728 metadata, fetched pre-auth). Auth on the endpoint is still enforced by McpAuthGuard.

server/mcp/tools/schemas.ts
thesocialdev: Fixed in 07aedb41. The create-tool inputs now derive from canonical Zod instead of re-declaring entity fields: reused the existing `PersonalitySchema` (now exported from server/interfaces/personality.interface.ts), and added colocated canonical schemas `server/source/schemas/source.zod.ts`, `server/claim/schemas/claim.zod.ts`, `server/verification-request/schemas/verification-request.zod.ts`. schemas.ts now derives each create schema via `.pick()`/`.extend()` (e.g. CreatePersonalitySchema = PersonalitySchema.pick({name,description,wikidata})), so entity fields are single-sourced. I left the list/get/pagination schemas local since they're query-param shapes, not entities. Note: the legacy create paths still use class-validator DTOs (per CLAUDE.md, not migrating those here) — these canonical Zod objects establish the single source the tools derive from without touching the DTOs.

thesocialdev:
## Security unification — MCP now uses the platform gateway (not a parallel path)

Per review feedback that the MCP auth/authorization was built separately (a breach for MCP agents rather than iterating on the platform), the branch now routes MCP through the **same** security primitives the rest of the app uses. Three parallel implementations were collapsed into one:

**Authentication — one bearer/identity gateway.** New `TokenIdentityService` (`server/auth/token-identity.service.ts`) centralizes Hydra introspection + Kratos identity resolution + the affiliation check + `request.user` shaping. All three guards now delegate to it:
- `M2MGuard`, `SessionGuard`, and `McpAuthGuard` no longer each carry their own introspection/affiliation/shaping.
- Side benefit: the platform M2M path now resolves **real user roles** for user tokens instead of hard-coding `integration`, and the affiliation check is unified onto a single fail-closed implementation (also hardens `SessionGuard`).

**Authorization — CASL, not a hand-rolled list.** MCP write tools were gating on a bespoke `WRITE_ROLES` array that disagreed with CASL (e.g. it treated an `integration` M2M token as create+update; CASL grants only Create). Replaced with `AbilityFactory.defineAbility(user, ns).can(Action.Create, "all")` — the platform's single authorization source. `WRITE_ROLES`/`resolveUserRole` deleted.

**Namespace — the platform's source of truth.** MCP inferred membership from the Ory identity role-map (`Object.keys(user.role)`); it now resolves membership via `NameSpaceService` against the `NameSpace` collection's `users[]`, the same source `NameSpaceGuard` uses. `getAllowedNamespaces` deleted.

**Verification:** full unit suite 292/292, MCP + representative controller e2e green, `build-ts` clean. Each step was TDD'd and adversarially reviewed; the guard refactors were checked for login-breaking risk (none — `SessionGuard` is behavior-preserving in any configured env; existing M2M client-credential integrations unchanged).

Design/plan: `docs/superpowers/plans/2026-07-19-mcp-security-unification.md`. One pre-existing cross-namespace gap on `add_review_comment`/`get_review_task` (acting on review tasks by id/data_hash without a per-namespace check — mirrors the existing web routes) is tracked as a separate follow-up, not introduced here.

thesocialdev:
**Restacked.** The platform authentication gateway (`TokenIdentityService` + `OryService.getIdentity` + the M2MGuard/SessionGuard unification) has been extracted into its own base PR #2560. This PR is now retargeted onto `feat/token-identity` and its diff shows only the MCP-specific work (17 commits): the MCP module/tools/routes, and MCP's own delegation to `TokenIdentityService` + CASL `AbilityFactory` write-gating + `NameSpaceService` membership. Review/merge #2560 first, then this. The branch was force-pushed for the restack (content is byte-identical to before — only re-parented onto the base); earlier review threads may show as outdated but their fixes are intact.

PR #2560, opened 2026-07-19
refactor(auth): unify bearer/session identity resolution into TokenIdentityService

thesocialdev:
# Description

Extracts the platform authentication gateway from the MCP server work (#2559) into its own PR. This unifies bearer/session identity resolution into a single service so the codebase has one security strategy instead of per-feature duplication.

**This is the base of a stack** — the remote MCP server PR (#2559) is retargeted on top of this branch.

Related Ticket # (issue) — _n/a_

## What changed

Previously, Hydra token introspection, Kratos identity resolution, the app-affiliation check, and `request.user` shaping were duplicated across `M2MGuard` and `SessionGuard` (and, in #2559, a third copy in the MCP guard). This introduces **`TokenIdentityService`** (`server/auth/token-identity.service.ts`) as the single resolver:

- `introspect(token)` — one Hydra introspection implementation.
- `isAffiliationValid(traits)` — one fail-closed affiliation check.
- `buildIdentityUser(traits, state)` / `shapeM2MUser(introspection)` — canonical `request.user` shaping.
- `resolveBearerToken(token)` — introspect → M2M shape, or resolve the real Kratos identity (affiliation + active-state checked); `null` = deny.

Guards now delegate:
- **`M2MGuard`** → `resolveBearerToken`. Behavior change (intended): a **user** bearer token now resolves to the user's real Kratos role instead of always being labeled `integration`. Client-credential tokens (`sub === client_id`) still resolve to the `integration` shape — existing integrations unchanged. A defensive try/catch keeps the guard from turning a resolver throw into a 500 (falls through to the session path).
- **`SessionGuard`** → uses `isAffiliationValid` + `buildIdentityUser`. Behavior-preserving in any configured environment; the affiliation check is unified onto the fail-closed implementation.

Also adds `OryService.getIdentity(id)` (Kratos admin identity lookup) which the resolver uses.

# Type of change

- [x] Existing feature enhancement (non-breaking change which modifies existing functionality)

# Testing

- `yarn test:unit` — new `TokenIdentityService` spec (10 tests: M2M/user resolution, inactive token, introspection/identity-lookup errors, affiliation fail-closed, inactive identity, shaping units); `M2MGuard` and `SessionGuard` specs updated and green; full `server/auth` suite 42/42.
- `yarn build-ts` clean.
- Adversarially reviewed for login-breaking risk: none for any correctly-configured environment (`app_affiliation` is a required deployment field); `SessionGuard` is behavior-preserving; no privilege escalation from the M2M change.

# Backend Changes
- [x] Endpoints remain secured — this refactors the shared guard internals; the global guard chain is unchanged.
- [x] New service has a defined interface (`AuthenticatedUser`, typed methods).

# Notes for reviewers

Design/rationale: `docs/superpowers/plans/2026-07-19-mcp-security-unification.md` (steps 1–4). The MCP server PR (#2559) builds on this and additionally routes its own guard through `TokenIdentityService` and gates its tools via the existing CASL `AbilityFactory` and `NameSpaceService`.

### DIFF
--- a/server/app.module.ts
+++ b/server/app.module.ts
@@ -61,6 +61,7 @@ import { FeatureFlagModule } from "./feature-flag/feature-flag.module";
 import { GroupModule } from "./group/group.module";
 import { SessionOrM2MGuard } from "./auth/m2m-or-session.guard";
 import { M2MGuard } from "./auth/m2m.guard";
+import { TokenIdentityModule } from "./auth/token-identity.module";
 import { CallbackDispatcherModule } from "./callback-dispatcher/callback-dispatcher.module";
 import { AiTaskModule } from "./ai-task/ai-task.module";
 import { TrackingModule } from "./tracking/tracking.module";
@@ -134,6 +135,7 @@ export class AppModule implements NestModule {
                 ViewModule,
                 SitemapModule,
                 OryModule,
+                TokenIdentityModule,
                 ReportModule,
                 CaptchaModule,
                 ImageModule,
--- a/server/auth/m2m.guard.spec.ts
+++ b/server/auth/m2m.guard.spec.ts
@@ -3,27 +3,12 @@ import { ExecutionContext } from "@nestjs/common";
 import { Reflector } from "@nestjs/core";
 import { ConfigService } from "@nestjs/config";
 import { M2MGuard } from "./m2m.guard";
+import { TokenIdentityService } from "./token-identity.service";
 import { mockAuthConfigService } from "../mocks/AuthMock";
-
-// Mock @ory/client. Variables referenced inside a vi.mock() factory must be
-// declared via vi.hoisted() because vi.mock() is hoisted above all imports
-// during the Vitest transform pipeline. Constructor mocks must use `function`
-// (not arrow functions) so `new Configuration(...)` works under Vitest 4.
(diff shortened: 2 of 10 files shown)


thesocialdev:
@caneppelevitor I want to prepare the platform for an MCP functionality, but that requires some changes in the auth flow.

caneppelevitor:
Made a small change: `resolveBearerToken` could throw instead of returning null, an identity without traits passes the affiliation check when app_affiliation is unset, then `buildIdentityUser` throws on `traits.user_id`. In #2559 the MCP guard calls this without a catch, that throw becomes a 500 instead of a 401. Fixed with a null guard, documented the never-throws + scope-blind contract, added 3 regression tests.

caneppelevitor:
> Made a small change: `resolveBearerToken` could throw instead of returning null, an identity without traits passes the affiliation check when app_affiliation is unset, then `buildIdentityUser` throws on `traits.user_id`. In #2559 the MCP guard calls this without a catch, that throw becomes a 500 instead of a 401. Fixed with a null guard, documented the never-throws + scope-blind contract, added 3 regression tests.

Saying that, I believe we could merge this base MR and then do another verification on the MCP one, as I was doing some testing and there may be some things to tighten up in the documentation

caneppelevitor (approved):

